PR #64, opened 2019-02-06
refactor get, update and create API

aabarata:


.ciconfig
```diff
@@ -11,7 +11,7 @@ pipeline {
           sh 'npm install'
         }
         dir('linkurious-server') {
-          git branch: 'develop', credentialsId: 'dd38fcea-d471-4856-a009-4029127c6450', url: 'git@github.com:Linkurious/linkurious-server.git'
+          git branch: 'query-ux', credentialsId: 'dd38fcea-d471-4856-a009-4029127c6450', url: 'git@github.com:Linkurious/linkurious-server.git'
```
maxime-allex: Dangerous, this should not be touched

package.json
```diff
@@ -82,6 +82,8 @@
     "pretty-quick": "^1.8.0"
   },
   "dependencies": {
-    "superagent": "^3.5.2"
+    "superagent": "^3.5.2",
+    "rxjs": "^6.3.3",
```
maxime-allex: remove rxjs from here and move it in shared from dev-dependencies to dependencies

.ciconfig
aabarata: Done!

package.json
aabarata: Done!

PR #66, opened 2019-03-13
API refactoring

maxime-allex:
Finally what has been done : 

- add the transformer and listener to the rest-client

- add a method `request` in `modules/module` that handle response transformation and dispatching error events

- I only touched new API

### WHY ?

Because, changing all API responses represent too much work in LKE, It can cause many regression and it will be really difficult too validate (some features had too much edge cases to be sure at 100% that everything works fine).

### SO WHAT ?

Only new APIs will use this new strategy of (transformer + listener), old ones will still work has before until we refactor them.

Only new APIs needs to be validated. It will be a bit harder too maintain but we limit regressions and possibly critical bugs.

### DIFF
--- a/src/http/fetcher.ts
+++ b/src/http/fetcher.ts
@@ -7,13 +7,23 @@
  * File:
  * Description :
  */
+
 'use strict';
 
-import { IHttpDriver, IHttpResponse, IFetchConfig, IDataToSend, IFetcherClientState } from '../../index';
+import {
+  IHttpDriver,
+  IHttpResponse,
+  IFetchConfig,
+  IDataToSend,
+  IFetcherClientState,
+  FetcherConfig,
+  IClientState,
+} from '../../index';
 import { LinkuriousError } from '../LinkuriousError';
 import { DefaultHttpDriver } from './DefaultHttpDriver';
 import { Logger } from '../log/Logger';
 import { Utils } from './utils';
+import { Tools } from 'linkurious-shared';
 
 export class Fetcher {
   private static SOURCE_KEY_TEMPLATE: string = '{dataSourceKey}';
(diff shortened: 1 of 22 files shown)


src/http/fetcher.ts
```diff
@@ -22,51 +32,27 @@ export class Fetcher {
...
+  private _clientState: IClientState;
   private readonly _baseApiURL: string;
 
-  constructor(logger: Logger, clientState: IFetcherClientState, baseUrl: string, httpDriver?: IHttpDriver) {
+  constructor(logger: Logger, clientState: IClientState, baseUrl: string, httpDriver?: IHttpDriver) {
     this._httpDriver = httpDriver ? httpDriver : new DefaultHttpDriver();
     this._logger = logger;
     this._clientState = clientState;
-    this._baseUrl = baseUrl.endsWith('/') ? baseUrl : (baseUrl + '/');
+    this._baseUrl = baseUrl.endsWith('/') ? baseUrl : baseUrl + '/';
```
aabarata: Be careful when merge to not accept this ```endsWith ``` over the previous fix

src/module/VisualizationParser.ts
```diff
@@ -118,4 +114,11 @@ export class VisualizationParser {
       edges: Array.from(me.values()),
     };
   }
+
+  /**
+   * Return true if an item is a node
+   */
+  public static isNode(item: IItem): item is INode {
```
aabarata: If the rest client has linkurious-shared as dependency why not use ```Tools.isNode() ``` ?

src/http/fetcher.ts
maxime-allex: endsWith was not the problem. The problem on IE was about the baseURI API. ButI will take care of it.

src/module/VisualizationParser.ts
maxime-allex: Actually is method is not used anymore :) I removed it

PR #81, opened 2019-05-16
Mock APIs

francescoinfante:


src/http/fetcher.ts
```diff
@@ -181,6 +163,11 @@ export class Fetcher {
     for (const key of Object.keys(pathParams)) {
       url = url.replace(`{${key}}`, encodeURIComponent(pathParams[key]));
```
Diachk: I forgot to skip this when the value is undefined

src/models/Alert.ts
```diff
@@ -0,0 +1,46 @@
...
+// TS2019-DONE
+
+import {FolderChildren, IDataSourceParams, IPersistedItem} from './Model';
+
+export interface ICreateAlertFolderParams extends IDataSourceParams {
+  title: string;
+  parent?: number;
+}
+
+export interface ICreateAlertFolderResponse extends IPersistedItem {
```
Diachk: why do we need the sourceKey in the response ?

test/index_spec.ts
```diff
@@ -416,64 +413,6 @@ describe('Linkurious class', () => {
     });
   });
 
-  describe('getEdgeTypes method', function() {
```
Diachk: why remove these tests ?

francescoinfante: we decided to not write tests anymore in the rest client except for rest client specific features (converting LkNode to Ogma node).

instead we will use the rest client directly to invoke API in the server tests

src/models/Alert.ts
francescoinfante: we don't indeed

src/http/fetcher.ts
francescoinfante: I did that, sourceKey is injected automatically if undefined in path.

test/index_spec.ts
Diachk: don't you think we should at least check that we hit the right api and get the expected response, I think the objectives of the tests are different here, [no strong opinion about it]

francescoinfante: If we include the rest-client as part of the integration tests, we are going to test that we hit the right API and get the response.

I would like to avoid writing duplicate tests. Unsure what else to test specifically here.

PR #101, opened 2019-09-02
Add data-source configuration options

Diachk:


src/models/DataSource.ts
```diff
@@ -0,0 +1,217 @@
...
+  canIndexEdges: boolean;
+  searchHitsCount: boolean;
+  alternativeIds: GenericObject<string>;
+  maxNodeCategories?: number;
+  externalIndex: boolean;
+  canCount: boolean;
+  dialects: string[];
+}
+
+export interface ISpecialProperties {
```
maxime-allex: What is this ?

src/models/DataSource.ts
```diff
@@ -0,0 +1,217 @@
+/**
+ * LINKURIOUS CONFIDENTIAL
+ * Copyright Linkurious SAS 2012 - 2019
+ *
+ * - Created on 2019-09-02.
+ */
+import { GenericObject, ICaptionsConfig } from './Configuration';
```
maxime-allex: Why using interfaces and not types ?

src/models/DataSource.ts
```diff
@@ -0,0 +1,217 @@
...
+  key: string;
+  configIndex: number;
+}
+
+export interface IDataSourceFeatures {
+  immutableNodeCategories: boolean;
+  canCountBeforeIndexation: boolean;
+  canIndexEdges: boolean;
+  searchHitsCount: boolean;
+  alternativeIds: GenericObject<string>;
```
maxime-allex: Since you type `alternativeIds` below, why not using it there ?

src/models/DataSource.ts
Diachk: What would be the benefit here?

maxime-allex: Just consistency and maintenability. Interfaces should be used for classes.

francescoinfante: when possible we use interfaces, because they can be used with implements and extends keywords and they work exactly as a `type` for the rest.

we define it a `type` only when needed for more complex types that couldn't be phrased as an interface and therefore cannot be used with the implements and extends keywords.

francescoinfante: @maxime-allex @Diachk open to discuss, but I don't see any gain in transforming an interface into a type.

maxime-allex: You can extends a type to by using `type Type = OtherType + {...}`.

For `implement`, I agree, but you use that keyword with a Class.

That's just a question of taste. I feel more confortable when usages are simple and clearly defined.

Anyway, that's your repo so do whatever you want 😛

src/models/DataSource.ts
Diachk: let's ask @francescoinfante, `sources.settings.specialProperties`   -->     Properties with special access rights for this source

src/models/DataSource.ts
Diachk: CaptionConfig is a class in Linkurious server that implements ICaptionConfig, so it makes sense for us to define it as an interface. Except maybe code style or personal preference, i don't see additional benefit as well

maxime-allex: > CaptionConfig is a class in Linkurious server that implements ICaptionConfig, so it makes sense for us to define it as an interface

Yes

> Except maybe code style or personal preference

Yes, that's what I said below.

francescoinfante:
Done a post review
f8a7b4a

Missing to review the configuration model, everything else is good :)

@maxime-allex @Diachk

Diachk:
Good for me

PR #104, opened 2019-09-12
Property Key Access Rights

francescoinfante:
WIP

### DIFF
--- a/src/models/AccessRights.ts
+++ b/src/models/AccessRights.ts
@@ -24,13 +24,13 @@ export interface ISetAccessRightsParams extends IDataSourceParams {
   groupId: number;
   type: AccessRightType;
   targetType: TargetType;
-  targetItemType?: string;
-  targetName: string;
+  targetItemType?: string; // e.g.: 'Person'
+  targetName: string; // e.g.: 'Person' if targetType is 'nodeCategory' or 'name' if targetType is 'nodePropertyKey'
 }
 
 // Examples:
 //
-// NODE_PROPERTY_KEY access right type:
+// NODE_PROPERTY_KEY access right:
 // {
 //   groupId: 10,
 //   type: 'read',
@@ -39,7 +39,7 @@ export interface ISetAccessRightsParams extends IDataSourceParams {
 //   targetName: 'name'
 // }
 //
-// NODE_CATEGORY access right type:
+// NODE_CATEGORY access right:
 // {
 //   groupId: 10,
 //   type: 'read',
--- a/src/models/AccessRights.ts
+++ b/src/models/AccessRights.ts
@@ -21,11 +21,48 @@ export enum TargetType {
 }
 
 export interface ISetAccessRightsParams extends IDataSourceParams {
-  groupId: number;
-  type: AccessRightType;
-  targetType: TargetType;
-  targetItemType?: string; // e.g.: 'Person'
-  targetName: string; // e.g.: 'Person' if targetType is 'nodeCategory' or 'name' if targetType is 'nodePropertyKey'
+  groupId: number; // e.g.: 123
+  type: AccessRightType; // e.g.: 'read'
+  target:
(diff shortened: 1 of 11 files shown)


src/models/AccessRights.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright Linkurious SAS 2012 - 2019
+ *
+ * - Created on 2019-09-12.
+ */
+
+// TS2019-DONE
+
+import {IDataSourceParams} from './DataSource';
+
+export interface ISetAccessRightsParams extends IDataSourceParams {
```
davidrapin: A comment with example values for each field would be welcome, otherwise it's difficult to guess what each field should be for a user.

src/models/AccessRights.ts
```diff
@@ -0,0 +1,22 @@
...
+  groupId: number;
+  type: string;
```
davidrapin: should be a union type `type AccesRightType = "read" | "write" | "edit" | "do" | "none"`

src/models/AccessRights.ts
```diff
@@ -0,0 +1,22 @@
...
+  targetType: string;
```
davidrapin: should be a union type of `"nodeCategory","edgeType","alert","action"`

src/models/AccessRights.ts
```diff
@@ -0,0 +1,22 @@
...
+  targetItemType?: string;
```
davidrapin: what is this for?

src/models/Schema.ts
```diff
@@ -115,27 +119,41 @@ export interface IGraphSchema {
...
+export enum TypeAccessLevel {
+  READABLE = 'readable',
+  EDITABLE = 'editable',
+  WRITABLE = 'writable'
+}
+
+export enum PropertyAccessLevel {
+  READABLE = 'readable',
+  EDITABLE = 'editable',
+  WRITABLE = 'writable'
```
davidrapin: there is no "writable" for properties, it's either no-access, read or edit.

src/models/Schema.ts
```diff
@@ -115,27 +119,41 @@ export interface IGraphSchema {
   results: IGraphSchemaType[];
 }
 
-export enum AccessLevel {
+export enum AnyTypeAccessLevel {
```
davidrapin: why the "any" prefix? I'm not sure what it is supposed to mean here.

francescoinfante: it's used for the access right of the "any" category for builtin groups. It's the right to create nodes with new categories when schema is partial mode.

Normally the accessLevel is always greater than "none" because "none" is filtered out except that it can be "none" for this particular case.

src/models/AccessRights.ts
francescoinfante: to set the itemType (e.g. 'COMPANY') to which an access right is referring too when targetType is 'nodePropertyKey'.

src/models/Schema.ts
davidrapin: Ok

src/models/AccessRights.ts
davidrapin: this this should be an union type as well, like `"nodeCategory", "edgeType", "nodePropertyKey", "edgePropertyKey"`

src/models/Schema.ts
davidrapin: A comment in the code would be welcome :)

src/models/AccessRights.ts
francescoinfante: that's targetType and not targetItemType.
Calling it targetLabel would have been clearer, but we prefered to rename "label" to "itemType".

davidrapin: Ok, now I understand, so it's `targetType` that should be union type.

BTW adding a example of value for each field that is not a union type in an inline comment in the interface could be helpful to get an idea of that these fields are.

PR #147, opened 2020-04-10
POC: alternative way to write requests

pastahito:


davidrapin:
@pastahito could you explain what this is trying to solve? It's difficult to understand the intent without context.

francescoinfante:
@davidrapin my fault that @pastahito didn't create an issue for this, he was waiting for me to review it to see if it was worth an issue :)

He found a way to simplify the way we write rest-client methods, in particular how we declare that methods can return expected errors.

I will validate if this change works and let Renato create an issue for mahmoud (since it will be a good Typescript exercise for him)

davidrapin:
@francescoinfante thank you for the clarification! I kinda understood the intent from reading the code, but was interested in specifically how the alternative way was mean to solve the issue.

pastahito:
@davidrapin @francescoinfante I just added the issue as an exercise for @mzorkany

pastahito:
@davidrapin the issue was that since the refactoring of RC, we used `this.handle(...).request(...)`, but we wanted to use directly `this.request(...)`. The reason why we've been using `handle(...)` was because in TS you can specify either none or all the generics of a function, and we wanted to specify some of them and let the others to be inferred. So `handle()` was used as a workaround to get the inferred generics, and request the explicit ones. But it always felt weird to use js to patch ts limitations (https://github.com/microsoft/TypeScript/issues/10571). 

The alternative way is about using `this` (type of Request) to hold the explicit type that is currently in `request()`, and make `request()` to infer the types that are currently inferred by `handle()`.

@francescoinfante correct me if I'm missing anything

PR #149, opened 2020-04-23
Expose plugin interfaces

Diachk:


package.json
```diff
@@ -43,6 +43,7 @@
     "typescript": "3.8.3"
   },
   "dependencies": {
+    "@types/express": "4.17.3",
```
pastahito, line 46: it's better as dev dependecy

francescoinfante: @pastahito actually, I'm not sure. if people write plugins they need the type

Diachk: yes, the aim was to expose interface `PluginRouteOptions` to the developer of the plugin

PR #155, opened 2020-06-26
LKE-446: Bootstrap new ci

enys:


enys:
Merging this commit will imply moving away from the old ci to the new CI

enys:
@davidrapin @francescoinfante @maxime-allex  how would you like to move forward on this one ?

davidrapin:
@enys 
- clarify release needs with backend team
  - do we sometimes need to release for older branches (e.g. 2.8.x?) 
  - ?
- clarify needs for development version
  - how to get rid of dependencies from other projects to the github repo of the rest client directly?) => use `~` or `^` prefix in dev, prevent these prefixes in prod.
  - ?
- merge this PR

thoughts?

enys:
This would make sense for linkurious-shared.
Here the CI does not push an asset. So it is not needed.

davidrapin:
@enys I think the CI should push assets for this repo (i.e. publish to npm)

enys:
I wasn't precise. Old co didn't push any assets at all.
New ci already is configured to publish to npm

davidrapin:
@enys Since this one pushes to NPM, I think we will have to discuss what we discussed yesterday (how to handle dependencies to non-public dev versions) with developers.

Which implies:
- agree on how to handle development builds in this lib (dependency to github vs. dependency to a tag in nexus)
- implement the required changes in this lib + its dependencies (e.g. update the frontend)

Do you want to merge this PR before going forward with this?

francescoinfante:
before merging anything, we need a presentation of the new CI :D

package.json
```diff
@@ -20,8 +20,12 @@
   "scripts": {
     "prepare": "tsc",
     "build": "tsc",
+    "test": "npm run compile && mocha dist/test",
+    "test:unit": "nyc --reporter=text-summary --reporter=cobertura --report-dir=reports  mocha --reporter mocha-junit-reporter --reporter-options mochaFile=./reports/unit/junit-test-results.xml dist/test",
+    "compile": "tsc -b",
```
francescoinfante, line 25: can you explain the build flag?

.gitignore
```diff
@@ -3,6 +3,11 @@ node_modules
 
 # IDEA config
 .idea
+*.swp
```
francescoinfante, line 6: what is it?

davidrapin:
@enys I think it's time to:
- document the goals (e.g. make PRs safer to merge, run the linter on each commit, only depend on built assets, never release manually), etc.
- document the "how"
- give a short presentation with slides mayb?

enys:
@davidrapin yes !

package.json
enys: rather than my bad explanation : https://www.typescriptlang.org/docs/handbook/project-references.html

TLDR: faster ( incremental builds)

.gitignore
enys: this is a gobal discussion to have. Either remove all editor files, or have them in each project, with a templated .gitignore.

Specificaly vim here.

enys:
> Do you want to merge this PR before going forward with this?

I think it should be the maintainers choice. However : there is no impact on current workflow, so I would merge.

package.json
francescoinfante: we already get incremental builds by tsconfig

.gitignore
francescoinfante: just add a # VIM comment for now :D

enys: a bit like idea config for idea .... :D

package.json
enys: which isn't present on all projects sadly

enys: it is not limited to incremental builds.

enys:
@francescoinfante @davidrapin @aminerhanemi all good to go on this one

Jenkinsfile
```diff
@@ -0,0 +1,8 @@
+@Library('linkurious-shared')_
```
francescoinfante, line 1: @enys what does this mean?
rest-client doesn't depend on shared and shared is a private repo :)

enys:
This is not related with nodeJs. This is loading the shared libraries for
Jenkins. Repository named Jenkins-shared-libraries

Jenkinsfile
enys: https://www.jenkins.io/doc/book/pipeline/shared-libraries/

francescoinfante: calling it linkurious-shared is going to confuse everyone, can we rename it to linkurious-jenkins-shared? (or linkurious-jenkins-common)

enys: I don't think we should prefix, in a specific context. ie would it make sense to call linkurious-shared linkurious-shared-node ? Here you are specificaly in a Jenkinsfile, It does not look at all like a nodeJs import.

enys:
Lets merge this today, before I have to rebase for the 4th time :)

Issue #156, opened 2020-07-03
Configure group styles of all edge groups


mzorkany:
As an Admin I can customize the style of all EG for a DS.

Changes in the Rest Client:
Create interface `DataSourceDefaultStyles` extends `Styles`
`interface DataSourceDefaultStyles extends Styles {edgeGroup: EdgeStyle}`

francescoinfante:
Done: https://linkurious.atlassian.net/browse/LKE-148

PR #157, opened 2020-07-03
#156 edge grouping style configuration

mzorkany:


src/api/displayTypes.ts
```diff
@@ -132,6 +132,10 @@ export interface Styles {
   edge: Array<IStyleRule<EdgeStyle>>;
 }
 
+export interface DefaultStyles extends Styles {
+  edgeGroup: EdgeStyle;
+}
+
```
Diachk, line 138: Why not add `edgeGroup` directly to the `Styles` interface ?

Diachk: Actually I understand this know, after reviewing the backend code.
Could you rename it to `DataSourceDefaultStyles` though to be consistent with the backend code
or any other name, as long as it is used and understood consistently everywhere

PR #182, opened 2020-09-21
Sequelize v6

pastahito:
This is the RC used by this server PR: Linkurious/linkurious-server#2247


pastahito:
Nothing to merge because the bug found in 2.9.1 was fixed after 2.9.7, but it still shows the merge button 🤔...

PR #189, opened 2020-11-05
LKE-696 Update INeo4jConfig interface

mzorkany:


francescoinfante:
ok for me (i'm adding also databaseName because it' missing :P )

PR #193, opened 2020-11-20
[WIP] feat: LKE-2157 add header to track use of API

EricMallard:


davidrapin:
@EricMallard did you intentionnalty skip merging? If so, do you intend to come back to this is the future? If not, I would be in favor of deleting the branch to avoid accumulating dead branches.

EricMallard:
@davidrapin actually I skipped merging on purpose since it should come back. The idea was avoiding polluting the PR opened. However, I agree it might pollute the branches and without some context, it might end up being confusing. @francescoinfante  how probable is it that it will come back at some point? Should I delete the branch (the work done was little, so I don't see a big loss in deleting)?

francescoinfante:
@EricMallard let's delete, it's 4 lines :)

also fyi, you may delete the branch, but the PR even if closed, remember the last commit and therefore the code

PR #207, opened 2021-01-12
LKE-2345 Comments on matches

pastahito:
EPIC: https://linkurious.atlassian.net/browse/LKE-2345

Server PR: Linkurious/linkurious-server#2558
Shared PR: https://github.com/Linkurious/linkurious-shared/pull/140


src/api/Alert/index.ts
```diff
@@ -195,9 +208,62 @@ export class AlertAPI extends Request {
...
+      params: params
+    });
+  }
+
+  /**
+   * Get the last created action of a match if any.
+   */
+  public async getLastMatchAction(
+    params: Pick<IGetMatchActionsParams, 'alertId' | 'matchId' | 'sourceKey'>
+  ): Promise<
```
Diachk, line 239: What is the objective of this method ?
I feel there might be something wrong here but not sure.
If the objective is to the the last comment on a status change, there might be a bug if the last action is a comment and not a status change.
If the objective is to get the last comment regardless of the action then why have a specific method for that when  `getMatchActions` already exists.

src/api/Alert/index.ts
```diff
@@ -195,9 +208,62 @@ export class AlertAPI extends Request {
   }
 
   /**
-   * Do an action (open, dismiss, confirm, unconfirm) on a match.
+   * Update a comment on a match if the user that triggered the update is the author
```
Diachk, line 211: we should add a dot at the end of line for consistency with the other descriptions

src/api/Alert/index.ts
```diff
@@ -195,9 +208,62 @@ export class AlertAPI extends Request {
...
+    return this.request({
+      errors: [FEATURE_DISABLED, UNAUTHORIZED, DATA_SOURCE_UNAVAILABLE, FORBIDDEN, NOT_FOUND],
+      url: '/:sourceKey/alert/match/comment/:commentId',
+      method: 'PATCH',
+      params: params
+    });
+  }
+
+  /**
+   * Delete a comment on a match if the user that triggered the deletion is the author
```
Diachk, line 223: here as well we should add a dot at the end of the line to be consistent.

Diachk (changes requested):
Review done.
Here as well you should change the base to develop-next

src/api/Alert/index.ts
pastahito: Indeed the frontend can just use the `getMatchActions`, this can be negotiated.

@maxime-allex are you ok if we remove `getLastMatchAction` method from the RC? You just need to update this:
```
  public async getLastAction(matchId: number, alertId: number): Promise<MatchAction> {
    const action = await this._restClient.alert.getLastMatchAction({
      matchId: matchId,
      alertId: alertId
    });
    if (action.isSuccess()) {
      return action.body;
    }
  }
  ```
To this:
```
  public async getLastAction(matchId: number, alertId: number): Promise<MatchAction> {
    const action = await this._restClient.alert.getMatchActions({
      matchId: matchId,
      alertId: alertId,
      action: [
        MatchActionType.CONFIRM,
        MatchActionType.DISMISS,
        MatchActionType.OPEN,
        MatchActionType.UNCONFIRM
      ],
      offset: 0,
      limit: 1
    });
    if (action.isSuccess()) {
      return action.body.matchActions[0];
    }
  }
```

francescoinfante:
@pastahito why it target develop and not develop-workflow?

src/api/Alert/index.ts
Diachk: @maxime-allex are you ok with this ? We are about to start merging everything

Diachk (approved):
LGTM

src/api/Alert/index.ts
maxime-allex: Of course you can remove it. Seeing the last action is not part of the scope anymore... 🥰

PR #212, opened 2021-01-22
LKE-2577 Cosmos idOnly option

mzorkany:


yigitakbulut (approved):

src/api/GraphQuery/types.ts
```diff
@@ -66,12 +66,14 @@ export interface EnumTemplate extends TemplateField<TemplateFieldType.ENUM> {
 export interface NodeTemplate extends TemplateField<TemplateFieldType.NODE> {
   options?: {
     categories?: string[];
+    serialize?: string;
```
pastahito, line 69: From what I understood in the ticket's last comment https://linkurious.atlassian.net/browse/LKE-2577?focusedCommentId=15570, the specs is `{idOnly?: boolean}`. What was the final decision?

mzorkany: @pastahito 

The original specs is to have a `serialize` option which accepts only the value `idOnly` as a string

The last comment in the ticket was from a discussion that i had with @davidrapin and @Diachk regarding the expected output when the `serialize` option is passed in the query template

davidrapin: @mzorkany can you point me to spec you are refering to?

mzorkany: @davidrapin 
https://linkurious.atlassian.net/browse/LKE-2577?focusedCommentId=15570

its written in the description of the ticket

pastahito: Nothing to changed, clarified by https://linkurious.atlassian.net/browse/LKE-2577?focusedCommentId=15927

pastahito (approved):

PR #217, opened 2021-02-09
LKE-155 query owner details

mzorkany:


yigitakbulut (approved):

src/api/GraphQuery/types.ts
```diff
@@ -200,6 +200,10 @@ export interface GraphQuery {
   dialect: GraphQueryDialect;
   description: string;
   sharing: GraphQuerySharingMode;
+  owner?: {
```
francescoinfante, line 203: @mzorkany @yigitakbulut why optional?

yigitakbulut: @francescoinfante the built in queries have no owner, so there are two options, we either null or something like that for built in queries or make the owner object optional. What do you think?

francescoinfante: @mzorkany @yigitakbulut make 100% sense :) (in Rest client it's good practice to write why something can be undefined as a comment)

PR #219, opened 2021-02-11
LKE-2756 configure indexed properties

mzorkany:


src/api/GraphSchema/types.ts
```diff
@@ -9,6 +9,7 @@ import {ItemTypeAccessRightType, PropertyAccessRightType} from '../AccessRight';
 
 export interface IStartSchemaSamplingParams extends IDataSourceParams {
   reset?: boolean;
+  defaultVisibility: DataVisibility;
```
francescoinfante, line 12: @mzorkany this should be optional

src/api/GraphSchema/types.ts
```diff
@@ -129,6 +130,7 @@ export interface GraphSchemaProperty {
   required: boolean;
   visibility: DataVisibility;
   propertyType: PropertyType;
+  indexed: boolean | null;
```
francescoinfante, line 133: @mzorkany don't use null, indexed is always defined for the frontend.

and if it wasn't in some cases, you should use indexed?: boolean with a comment explainin why.

src/api/DataSource/types.ts
```diff
@@ -37,6 +37,8 @@ export interface DataSourceFeatures {
   alerts: boolean;
   canDryRun: boolean;
   supportNativeDate: boolean;
+  searchEnabled: boolean;
+  isConsistent: boolean;
```
francescoinfante, line 41: here we are in "DataSourceFeatures". many things can be consistent in this scope. Name accordingly to the subject that it's "DataSourceFeatures".

francescoinfante (changes requested):

PR #220, opened 2021-02-12
remove timeline flag from config type

EricMallard:


yigitakbulut (approved):
That was actually for me to do, thanks for spotting it, otherwise it was going to be forgotten :)

EricMallard:
> That was actually for me to do, thanks for spotting it, otherwise it was going to be forgotten :)

teamwork 🦾

PR #245, opened 2021-04-29
LKE-3396 merged ES1 and ES2 config

francescoinfante:


src/api/Config/types.ts
```diff
@@ -208,29 +208,23 @@ export interface INeo4jSearchConfig extends IVendorConfig {
...
   timestampPropertyName?: string;
-}
-
-export interface IElasticSearchConfig extends ICommonElasticSearchConfig {
-  indexName?: string; // Deprecated option to be removed, kept here for backward compatibility
-}
-
-export interface IElasticSearch2Config extends ICommonElasticSearchConfig {
-  forceStringMapping?: string[];
-  caCert?: string;
```
mzorkany, line 233: what is exactly the `caCert` config and why its removed ?

francescoinfante: it's an old version of advanced.extraCertificateAuthorities that was working only with Elasticsearch.
Now it works for every external service.
No comment on the main PR? https://github.com/Linkurious/linkurious-server/pull/2668

mzorkany: @francescoinfante I haven't reviewed it yet but probably i will start reviewing it today :)

PR #247, opened 2021-04-30
LKE-3489 remove janusgraph

francescoinfante:


mzorkany (commented):
Since we are dropping JanusGraph, shouldn't we remove the graph & index files for janus graph as well ?

francescoinfante:
@mzorkany we are: https://github.com/Linkurious/linkurious-server/pull/2705

PR #248, opened 2021-05-10
LKE-3541-Fix-and-refactor-CM

yigitakbulut:
[ci:upload]

### DIFF
--- a/src/api/Alert/types.ts
+++ b/src/api/Alert/types.ts
@@ -4,12 +4,7 @@
  * - Created on 2019-10-30.
  */
 
-import {
-  IDataSourceParams,
-  IGetSubGraphParams,
-  PersistedItem,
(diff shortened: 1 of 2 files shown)


src/api/Alert/types.ts
```diff
@@ -35,7 +36,7 @@ export interface IPopulatedCaseVisualization extends ICaseVisualization {
   edges: VizEdge[];
 }
 
-export interface ICaseVisualization {
```
francescoinfante, line 38: Why use ICaseVisualization and not CaseVisualization?

also what's the point of extending a partial interface? everything is optional. The idea is to create a common interface both extends.

francescoinfante (changes requested):

francescoinfante (approved):

PR #252, opened 2021-06-02
LKE-2558-Custom-Actions-share-with-groups

yigitakbulut:


src/api/CustomAction/types.ts
```diff
@@ -58,7 +58,8 @@ export interface ParsedCustomAction {
 
 export enum CustomActionSharingMode {
```
EricMallard, line 59: Since they are now the same can we reuse, QuerySharingMode?

EricMallard: instead of creating 2 enum that are the same

EricMallard (changes requested):

src/api/GraphQuery/types.ts
```diff
@@ -223,13 +215,11 @@ export interface IGetQueriesParams extends IDataSourceParams {
   type: GraphQueryType;
 }
 
-export interface ICreateQueryParams extends IDataSourceParams {
+export interface ICreateQueryParams extends IDataSourceParams, SharingOptions {
```
EricMallard, line 218: @yigitakbulut I'm not sure we understood each other. For me it should not extend SharingOptions, it should have a property `sharingOptions` of type `SharingOptions`

PR #258, opened 2021-06-16
Bulk Case assignment

Diachk:
## DO NOT MERGE
Work in progress
**JIRA**  https://linkurious.atlassian.net/browse/LKE-3739

### DIFF
--- a/src/api/Alert/index.ts
+++ b/src/api/Alert/index.ts
@@ -30,7 +30,8 @@ import {
   IDeleteCaseCommentParams,
   IGetCaseActionsResponse,
   IUpdateCaseParams,
-  PopulatedCase
+  PopulatedCase,
+  IAssignCasesParams
 } from './types';
 
 export * from './types';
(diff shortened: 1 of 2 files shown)


src/api/Alert/index.ts
```diff
@@ -189,6 +190,18 @@ export class AlertAPI extends Request {
     });
   }
 
+  /**
+   * Assign one or more cases to a user.
+   */
+  public assignCases(params: IAssignCasesParams) {
```
yigitakbulut, line 196: I think we are missing the return type for this function here

yigitakbulut (changes requested):
Good for me except for the API definition

src/api/Alert/index.ts
Diachk: @yigitakbulut Not sure who decided this but no function in the rest-client has a return type you can check

yigitakbulut: I think the pattern is something like this:
`
public assignCases(this: Request<void>, params: IAssignCaseParams)
`

However since we return void, I agree that this can be discarded

Diachk: Indeed 👍

yigitakbulut (approved):
LGTM!!!

PR #303, opened 2021-09-07
LKE-3783: add optional owner to CustomAction type

maxime-allex:


src/api/CustomAction/types.ts
```diff
@@ -109,6 +109,10 @@ export interface CustomAction extends PersistedItem, SharingOptions {
   description: string;
   parsedTemplate: ParsedCustomAction;
   right: CustomActionRight;
+  owner?: {
```
francescoinfante, line 112: @maxime-allex why optional? you should add a comment in which case it's optional (or make it mandatory)

francescoinfante (changes requested):

Diachk (approved):
👍

src/api/CustomAction/types.ts
maxime-allex: I thought we wanted to return `owner` only if the custom action was shared (GraphQuery interface misled me).

francescoinfante (approved):

PR #318, opened 2021-11-17
LKE-4395 Properties order

Diachk:
https://linkurious.atlassian.net/browse/LKE-4395

### DIFF
--- a/src/api/DataSource/types.ts
+++ b/src/api/DataSource/types.ts
@@ -87,6 +87,7 @@ export interface DataSourceUserInfo {
 export interface ISetDefaultSourceStylesParams extends IDataSourceParams {
   styles?: IDataSourceDefaultStyles;
   captions?: ICaptions;
+  propertiesOrder?: IPropertiesOrder;
 }
 
 export interface IResetSourceStylesParams extends IDataSourceParams {
--- a/src/api/DataSource/types.ts
+++ b/src/api/DataSource/types.ts
@@ -4,7 +4,7 @@
  * - Created on 2019-10-25.
  */
 
-import { ICaptions, IDataSourceDefaultStyles, IPropertiesOrder } from '../displayTypes';
+import {ICaptions, IDataSourceDefaultStyles, IPropertiesOrder} from '../displayTypes';
 import {IDataSourceParams} from '../commonTypes';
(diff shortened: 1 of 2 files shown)


src/api/DataSource/types.ts
```diff
@@ -86,6 +87,7 @@ export interface DataSourceUserInfo {
 export interface ISetDefaultSourceStylesParams extends IDataSourceParams {
   styles?: IDataSourceDefaultStyles;
   captions?: ICaptions;
+  propertiesOrder?: IPropertiesOrder;
```
aminerhanemi, line 90: why propertiesOrder is needed here ?

Diachk: Isn't this how you set it from the UI using the setDefaultsAPI ?

aminerhanemi: you're right just the API name is confusing "setDefaultSourceStyles".
Now it is used for style caption and properties order

Diachk: You want to rename it to just `setDefaults` ?

src/api/DataSource/types.ts
```diff
@@ -78,6 +78,7 @@ export interface DataSourceUserInfo {
   features: DataSourceFeatures;
   defaultStyles?: IDataSourceDefaultStyles; // defined if withStyles or withCaptions was set to true in the request and the data-source is connected
   defaultCaptions?: ICaptions;
+  defaultPropertiesOrder?: IPropertiesOrder;
```
francescoinfante, line 81: when is it undefined?
what about "withStyles" and "withCaptions" params in IGetDataSourcesStatusParams interface?

francescoinfante (changes requested):

src/api/DataSource/types.ts
Diachk: It is defined when the properties order has been configured and undefined otherwise.

francescoinfante: why not always defined with default to empty object?
why not keeping consistent with the other styles?

Diachk: It is simpler to keep it defined with a default empty object

francescoinfante (approved):

PR #332, opened 2021-12-21
OPS-354: Move simple Node.js pipelines to k8s

enys:


yigitakbulut (approved):

enys:
@yigitakbulut any questions if you had a look at JSL ?

PR #335, opened 2021-12-28
LKE-2955 feature (datasource): persist sourceKey in configuration

gmasclet:
Data-source configuration has a `manualSourceKey` which is manually edited and override the computed `sourceKey` when set. This feature:
* Rename the `manualSourceKey` configuration field to `sourceKey`.
* Automatically set the `sourceKey` when successfully connecting to a datasource for the first time.

So persisting the `sourceKey` in configuration becomes the default behaviour and no longer requires manual action from the user.


yigitakbulut (approved):
LGTM

PR #349, opened 2022-02-18
LKE-5266: authentication required by default

gmasclet:
https://linkurious.atlassian.net/browse/LKE-5266

_As an Admin, I expect that only authorized users have access to my instance of LKE._

### DIFF
--- a/src/api/Auth/index.ts
+++ b/src/api/Auth/index.ts
@@ -9,9 +9,9 @@ import {Request} from '../../http/request';
 import {User} from '../User';
 
 import {
-  ICreateCurrentUserParams,
   ILoginOAuth2Params,
   ILoginParams,
+  ISetupAuthenticationParams,
   IUpdateCurrentUserParams
 } from './types';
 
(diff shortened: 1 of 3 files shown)


src/api/Auth/index.ts
```diff
@@ -78,6 +83,22 @@ export class AuthAPI extends Request {
     return response;
   }
 
+  /**
+   * Create the current user (in order to setup authentication).
+   */
+  public async createCurrentUser(this: Request<User>, params: ICreateCurrentUserParams) {
```
aminerhanemi, line 89: can we name it createFirstUser ? just a suggestion

aminerhanemi (changes requested):

src/api/Auth/types.ts
```diff
@@ -15,6 +15,12 @@ export interface ILoginOAuth2Params {
   state: string;
 }
 
+export interface ICreateCurrentUserParams {
```
aminerhanemi, line 18: same

aminerhanemi:
I added the ci:upload label so we can test it with the server and frontend

src/api/Auth/index.ts
gmasclet: I renamed it `setupAuthentication()` as agreed during our Zoom call.

src/api/Auth/types.ts
gmasclet: Renamed here too :ok_hand:

PR #350, opened 2022-02-22
LKE-5324- admin test smtp config (post review)

mzorkany:


src/api/mailer/index.ts
```diff
@@ -7,10 +7,10 @@ export class TestMailerConfigAPI extends Request {
   /**
```
yigitakbulut, line 7: Wdyt about changing class name to MailerAPI? @mzorkany

PR #353, opened 2022-02-28
LKE-5425: Count shared user assets

yigitakbulut:


src/api/User/types.ts
```diff
@@ -125,3 +125,7 @@ export interface IMergeUsersParams {
   from: number;
   to: number;
 }
+
+export interface ICountSharedUserAssetsParams {
+  id: number;
```
EricMallard, line 130: why id and not count?
also, should we have some additional data for the most shared user? with the name of the user that has received most assets + count of those assets?

src/api/User/types.ts
```diff
@@ -126,6 +126,18 @@ export interface IMergeUsersParams {
...
+
 export interface IGetAssetTransferEligibleUsersParams {
   id: number;
 }
+
+export interface ICountSharedAssets {
+  count: number;
+  user: {
+    id: number | null;
+    email: number | null;
```
EricMallard, line 141: missing the number of shared assets with this user, no?

PR #358, opened 2022-03-07
LKE-3820: Migrate assets of a deleted user

yigitakbulut:


src/api/User/index.ts
```diff
@@ -181,4 +182,16 @@ export class UserAPI extends Request {
       params: params
     });
   }
+
+  /**
+   * Get list of the users who have equivalent access rights to a given user
+   */
+  public getAssetTransferEligibleUsers(params: IGetAssetTransferEligibleUsersParams) {
```
EricMallard, line 189: are we missing the return type here?

PR #359, opened 2022-03-16
LKE-5654: Enforce license during installation

gmasclet:
https://linkurious.atlassian.net/browse/LKE-5654

_As an admin, I expect to install LKE for the first time with a valid license_

### DIFF
--- a/src/api/Config/types.ts
+++ b/src/api/Config/types.ts
@@ -22,7 +22,7 @@ export interface Configuration {
   ssoProvider?: 'oauth2' | 'saml2';
   url: string;
   setupAuthentication: boolean;
-  license?: {
+  license: {
     state: LicenseState;
   };
 
--- a/src/api/Config/types.ts
+++ b/src/api/Config/types.ts
@@ -21,6 +21,9 @@ export interface Configuration {
   ssoProvider?: 'oauth2' | 'saml2';
   url: string;
   setupAuthentication: boolean;
+  license?: {
+    state: 'valid' | 'expired' | 'missing';
+  };
 
   // partially available to not authenticated user
   access?: IAccessConfig;
@@ -28,9 +31,6 @@ export interface Configuration {
   // available to authenticated users
   advanced?: IAdvancedConfig;
   leaflet?: ILeafletConfig[];
-  license?: {
-    expired: boolean;
-  };
 
   // partially available to authenticated user
   alerts?: IAlertsConfig;
--- a/src/api/Config/types.ts
+++ b/src/api/Config/types.ts
@@ -7,6 +7,7 @@
 import {TlsOptions} from 'tls';
 
 import {GenericObject} from '../commonTypes';
+import {LicenseState} from '../License';
 import {IPluginConfig} from '../Plugin';
 import {OgmaNodeShape, OgmaEdgeShape} from '../displayTypes';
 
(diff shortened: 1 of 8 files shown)


src/api/Config/types.ts
```diff
@@ -21,16 +21,16 @@ export interface Configuration {
   ssoProvider?: 'oauth2' | 'saml2';
   url: string;
   setupAuthentication: boolean;
+  license?: {
+    state: 'valid' | 'expired' | 'missing';
```
aminerhanemi, line 25: can we use an enum for the type ?

src/api/License/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import {ISubmitLicenseParams, LicenseState} from './types';
+
+const {INVALID_LICENSE} = LkErrorKey;
+
+export class LicenseAPI extends Request {
+  /**
+   * Verify the supplied license and save it if it’s valid and there is no saved license yet.
+   */
+  public submitLicense(this: Request<LicenseState>, params: ISubmitLicenseParams) {
```
aminerhanemi, line 17: When this API will be called? I thought that the license will be saved when we call setupAuthentication and not the first call

src/api/License/types.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export enum LicenseState {
+  VALID = 'valid',
+  EXPIRED = 'expired',
+  MISSING = 'missing'
+}
+
+export interface SubmitLicenseResponse {
+  state: LicenseState.VALID | LicenseState.EXPIRED;
```
aminerhanemi, line 17: what happens when license is invalid ?

aminerhanemi: don't mind I saw it is returned as an error

src/api/Config/types.ts
gmasclet: sure, I've added a `LicenseState` enum

src/api/License/index.ts
gmasclet: This API has to be called if the license is missing (check the `license` value of the `GET /api/config` response). There are cases where the setupAuthentication endpoint won't be called (for instance if authentication is disabled); that's why this endpoint saves the uploaded license if there is no installed license yet.

If setupAuthentication is called without calling this endpoint first, it will also save the license with the same condition (internally, both endpoints call the same license service method). Submitting the license twice is not a big deal as it's a very small file, and the license checking/saving logic is idempotent.

src/api/License/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export * from './types';
+
+const {INVALID_LICENSE} = LkErrorKey;
+
+export class LicenseAPI extends Request {
+  /**
+   * Verify the supplied license and save it if it’s valid and there is no saved license yet.
+   */
+  public submitLicense(this: Request<SubmitLicenseResponse>, params: ISubmitLicenseParams) {
```
Diachk, line 19: To be consistent with my comments on the server PR, I think naming matter for readability and reusability.
Here we can create 2 aliases that do the same thing
`createLicenseIfMissing`
`verifyLicense`

open to discuss.

Diachk (dismissed):
Review done.
I agree with everything except the naming of the submitLicense api 👾

src/api/License/index.ts
gmasclet: I've renamed this method `saveLicenceIfMissing` as discussed together.

aminerhanemi (approved):

PR #377, opened 2022-03-29
LKE-5420 Unified Case List

mzorkany:


maxime-allex (changes requested):

src/api/Alert/types.ts
```diff
@@ -244,3 +244,39 @@ export type AlertPreview = Array<{
...
+export interface IFullCase {
+  alertName: string;
+  alertFolder: string;
+  alertDescription: string;
+  creationDate: Date;
+  status: CaseStatus;
+  statusChangedBy: Pick<User, 'id' | 'username' | 'email'> | null;
+  statusChangedOn: Date | null;
+  assignee: Pick<User, 'id' | 'username' | 'email'> | null;
+  attributes: ICaseColumns;
```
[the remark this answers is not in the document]
EricMallard: ICaseColumn? (without s)

src/api/Alert/types.ts
```diff
@@ -244,3 +244,39 @@ export type AlertPreview = Array<{
   edges: LkEdge[];
   columns: Array<string | number>;
 }>;
+
+export interface ICaseColumns {
```
maxime-allex, line 248: That should be `ICaseColumn` without an "s"

maxime-allex (changes requested):

maxime-allex (approved):

src/api/Alert/index.ts
```diff
@@ -321,4 +323,13 @@ export class AlertAPI extends Request {
       params: params
     });
   }
+
+  public getFullCaseList(this: Request<IFullCase[]>, params: IGetFullCaseListParams) {
+    return this.request({
+      errors: [FEATURE_DISABLED, UNAUTHORIZED, DATA_SOURCE_UNAVAILABLE, FORBIDDEN, NOT_FOUND],
```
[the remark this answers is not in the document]
maxime-allex: All these errors are handled automatically in the frontend.

src/api/Alert/types.ts
```diff
@@ -244,3 +244,46 @@ export type AlertPreview = Array<{
...
   columns: Array<string | number>;
 }>;
+
+export interface ICaseColumn {
+  type: AlertColumnType;
+  columnValue: string | number;
+  columnTitle: string;
+}
+
+export enum FullCaseListSortProperties {
```
EricMallard, line 254: @maxime-allex @mzorkany do you think we could align the FullCaseListSortProperties with the IFullCase property keys?
so for exemple we should not have assignedUser in one and assignee in the other. Also, createdAt and creationDate.
For me, it should be the same names.
(And incidentally, it makes it easier for me to do the sorting. If the user clicks on 1 column, I get the column key and send it to the backend. I do not need to map it to a FullCaseListSortProperties)
wdyt?
I would even go further and propose to make, when possible, the keys of the IFullCase interface one of the FullCaseListSortProperties value. In IFullCase:
`[FullCaseListSortProperties.CASE_ID]: number`
It makes it easier to maintain if those value change in the futur!

src/api/Alert/types.ts
```diff
@@ -244,3 +244,46 @@ export type AlertPreview = Array<{
...
+  assignedUser: Pick<User, 'id' | 'username' | 'email'> | null;
+  attributes: ICaseColumn[];
+}
+
+export interface IFullCaseListResponse {
+  totalCasesCount: number;
+  fullCaseList: IFullCase[];
+}
+
+export type fullCaseListSortBy = {by: FullCaseListSortProperties; direction: GetCasesSortDirection};
```
davidrapin, line 283: types should start with a capital letter

src/api/Alert/index.ts
```diff
@@ -321,4 +323,16 @@ export class AlertAPI extends Request {
...
+
+  /**
+   * Get all cases from alerts that a user has access to.
+   */
+  public getFullCaseList(this: Request<IFullCaseListResponse>, params: IGetFullCaseListParams) {
+    return this.request({
+      errors: [FEATURE_DISABLED, UNAUTHORIZED, DATA_SOURCE_UNAVAILABLE, FORBIDDEN, NOT_FOUND],
+      url: '/:sourceKey/alerts/cases/list',
+      method: 'GET',
+      params: params
```
davidrapin, line 335: where is the special encoding for params.sortBy handled?

mzorkany: @davidrapin 
Initially it was handled by the front-end  but now we figured that this might not be a good idea so we moved the encoding part to be handled by the res-client

src/api/Alert/index.ts
```diff
@@ -321,4 +323,17 @@ export class AlertAPI extends Request {
       params: params
     });
   }
+
+  /**
+   * Get all cases from alerts that a user has access to.
+   */
+  public getFullCaseList(this: Request<IFullCaseListResponse>, params: IGetFullCaseListParams) {
+    const encodedSortByParam = encodeURIComponent(JSON.stringify(params.sortBy));
```
davidrapin, line 331: you do not need to use encodeURIComponent, this part is handled by the HTTP Client itself
only JSON.stringify is needed

PR #394, opened 2022-04-22
LKE-5661: Update license

gmasclet:
https://linkurious.atlassian.net/browse/LKE-5661

_As an admin, I expect to update my expired license_


src/api/License/index.ts
```diff
@@ -6,25 +6,45 @@
...
       url: '/license',
       method: 'POST',
       params: params
     });
   }
+
+  /**
+   * Replace the saved license with the submitted one. Fails if submitted license is invalid.
+   */
+  public updateLicense(this: Request<SaveLicenseResponse>, params: ISaveLicenseParams) {
```
aminerhanemi, line 42: I think it is better to return LicenseInfo here

aminerhanemi (changes requested):

src/api/License/index.ts
aminerhanemi: not blocking (not sure if it is possible)

gmasclet: Discussed all together: OK to implement Amine's proposal, this avoids the frontend sending an additional request to retrieve the updated `endDate` value.

gmasclet: Done :ok_hand:

aminerhanemi (approved):

PR #395, opened 2022-04-25
LKE-4132: Reduce the amount of data exposed by the getUsers endpoint

yigitakbulut:


src/api/User/index.ts
```diff
@@ -61,7 +63,19 @@ export class UserAPI extends Request {
...
+      url: '/admin/users',
+      method: 'GET',
+      params: params
+    });
+  }
+
+  /**
+   * Get all the users or filter them by username, e-mail or group id.
+   */
+  public getUserList(this: Request<GetUserListResponse>, params: IGetUserListParams) {
```
davidrapin, line 75: what do you think about naming this one "searchUsers", and renaming the version for admins "searchUsersFull"?
That way, the names are consistent with the fact that both are for searching, but one returns simple users and the other one full users with more details.

PR #420, opened 2022-05-05
LKE-5770: Add interface for EmailNotifications in configuration

maxime-allex:


src/api/Config/types.ts
```diff
@@ -45,10 +45,28 @@ export interface Configuration {
   plugins?: IPluginConfig;
   dataSource?: SelectedDataSourceConfig;
   needRestart?: boolean;
+  emailNotifications: EmailNotifications;
 }
 
 export type DatabaseDialect = 'sqlite' | 'mysql' | 'mariadb' | 'mssql';
 
+export interface EmailNotifications {
```
yigitakbulut, line 53: This is not needed

src/api/Config/types.ts
```diff
@@ -45,10 +45,28 @@ export interface Configuration {
   plugins?: IPluginConfig;
   dataSource?: SelectedDataSourceConfig;
   needRestart?: boolean;
+  emailNotifications: EmailNotifications;
```
yigitakbulut, line 48: @maxime-allex If you scroll down to the bottom, we already have an interface for this called `IEmailNotificationsConfig`

maxime-allex: ok

yigitakbulut (approved):

PR #438, opened 2022-06-01
LKE-6264 case assignment notification config

mzorkany:


src/api/Config/types.ts
```diff
@@ -428,7 +428,8 @@ export interface ISMTPAuthConfig {
 
 export interface IEmailNotificationsConfig {
   alertNotifications: boolean;
-  notificationFrequency: string;
+  newCasesDigestFrequency: string;
```
gmasclet, line 431: For the sake of consistency, we may instead have:

```suggestion
  newCasesDigestNotificationFrequency: string;
```

So that both properties are named `xxxNotificationFrequency`. 

We could also change the other one to `caseAssignmentFrequency`, but this name is confusing without context.

Consider this comment as a non-blocking proposal.

gmasclet (approved):

PR #447, opened 2022-07-12
LKE-6144 feat: Add UCL preferences to user preferences

Mach12:


src/api/Config/types.ts
```diff
@@ -105,10 +106,18 @@ export interface IAuditTrailConfig {
   logPlugins?: boolean;
 }
 
+export interface IUnifiedCaseListPreferencesConfig {
+  alertIdsFilter?: number[];
+  caseStatusesFilter: CaseStatus[];
```
mzorkany, line 111: caseStatusesFilter should be optional

src/api/User/types.ts
```diff
@@ -7,11 +7,20 @@
 import {GenericObject, IDataSourceParams, PersistedItem} from '../commonTypes';
 import {AccessRight, Action, AnyAction} from '../AccessRight';
 import {IGuestPreferencesConfig} from '../Config';
+import {CaseStatus, FullCaseListSortBy} from '../Alert';
+
+export interface IUnifiedCaseListPreferences {
+  alertIdsFilter?: number[];
+  caseStatusesFilter: CaseStatus[];
```
mzorkany, line 14: caseStatusesFilter should be optional

src/api/Config/types.ts
```diff
@@ -105,10 +106,18 @@ export interface IAuditTrailConfig {
...
+  assignedUserIdsFilter?: number[];
+  sortBy: FullCaseListSortBy[];
+}
+
 export interface IUserPreferencesConfig {
   locale: string;
   pinOnDrag: boolean;
   incrementalLayout: boolean;
+  unifiedCaseListPreferences: IUnifiedCaseListPreferencesConfig;
```
mzorkany, line 120: No need for the re-declared interface `IUnifiedCaseListPreferencesConfig`
We can just use the one from `src/api/User/types`

mzorkany (changes requested):
Small changes required

src/api/User/types.ts
mzorkany: ahh sorry, no changes required here

mzorkany (changes requested):
Good job, few changes required

Also the new changes in `configUpdates.json` are missing

src/api/User/types.ts
```diff
@@ -7,11 +7,21 @@
...
+  assignedUserIdsFilter?: number[];
+  sortBy: FullCaseListSortBy[];
+  alertFoldersFilter?: number[];
```
EricMallard, line 17: could we call it folderIdsFilter (like alertIdsFilter, etc)

src/api/User/types.ts
EricMallard: why no changes required? I thought we said it would be optional?

src/api/User/types.ts
mzorkany: @EricMallard 
how about `alertFolderIdsFilter` ?
 just to distinguish between alert folders and visualisation folders

mzorkany (approved):

PR #470, opened 2022-08-04
LKE-6922 feat(space): Add create space API

aminerhanemi:


src/api/spaces/index.ts
```diff
@@ -0,0 +1,32 @@
...
+} = LkErrorKey;
+
+export class SpacesAPI extends Request {
+
+  /**
+   * Create a new space.
+   */
+  public createSpace(this: Request<ISpace>, params: ICreateSpaceParams) {
+    return this.request({
+      errors: [UNAUTHORIZED, FORBIDDEN, DATA_SOURCE_UNAVAILABLE, INVALID_PARAMETER],
```
aminerhanemi, line 25: Still need to check which errors are returned

gmasclet (changes requested):
This looks good to me beside a few misc changes :ok_hand:

src/api/spaces/index.ts
aminerhanemi: In which case the server should return INVALID_PARAMETER?

src/api/spaces/index.ts
```diff
@@ -0,0 +1,22 @@
+import {LkErrorKey} from '../../http/response';
```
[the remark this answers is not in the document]
aminerhanemi: I don't think it is needed in each file. the mention in the readme is enough. Maybe need to check with @enys .
I can add still add it

gmasclet (approved):

PR #482, opened 2022-08-10
LKE-6923 feat(spaces): Add getSpacesList API

aminerhanemi:


src/api/spaces/index.ts
```diff
@@ -19,4 +19,15 @@ export class SpacesAPI extends Request {
       params: params
     });
   }
+
+  /**
+   * Get spaces list.
+   */
+  public getSpacesList(this: Request<ISpace[]>) {
```
gmasclet, line 26: I suggest simplifying the method name to `getSpaces` and adding the source key parameter:

```suggestion
  public getSpaces(this: Request<ISpace[]>, params?: IDataSourceParams) {
```

src/api/spaces/index.ts
```diff
@@ -19,4 +19,15 @@ export class SpacesAPI extends Request {
       params: params
     });
   }
+
+  /**
+   * Get spaces list.
```
gmasclet, line 24: nit:

```suggestion
   * List spaces.
```

src/api/spaces/index.ts
aminerhanemi: what is the need of the sourcekey params ?

src/api/spaces/index.ts
aminerhanemi: I like starting with a verb to show it is an action

gmasclet (approved):
LGTM :+1:

PR #501, opened 2022-08-29
LKE-7035 create type definition for currency values

Diachk:


src/api/GraphSchema/types.ts
```diff
@@ -54,7 +54,8 @@ export enum PropertyTypeName {
   DATE = 'date',
   DATETIME = 'datetime',
   NUMBER = 'number',
-  STRING = 'string'
+  STRING = 'string',
+  CURRENCY = 'currency'
```
gmasclet, line 58: nit: conforming to the alphabetical ordering enhances readability, especially as the enum grows. We might consider moving the `CURRENCY` value between `BOOLEAN` and `DATE`

gmasclet (approved):
LGTM :+1:

src/api/GraphSchema/types.ts
Diachk: Thanks, done :)

PR #528, opened 2022-10-18
LKE-7223: Manage visualizations within a space

gmasclet:
https://linkurious.atlassian.net/browse/LKE-7223


aminerhanemi:
@Diachk  we still need to change getVisualizationTree

PR #557, opened 2023-01-25
LKE-7871 feat(UCL): add pagination offset in preferences

EricMallard:


src/api/Alert/types.ts
```diff
@@ -388,3 +389,8 @@ export interface IGetFullCaseListPreferencesResponse extends IFullCaseListPrefer
 export interface ISetFullCaseListPreferencesParams
   extends IDataSourceParams,
     IFullCaseListPreferences {}
+
+export const FULL_CASE_LIST_DEFAULT_SORTBY: FullCaseListSortBy = {
```
mzorkany, line 393: It doesn't make sense to define the const here as its only used by the server.

Usually on the back-end we declare these types of constants in the place where its going to be used,
In this case this const should be directly defined here
https://github.com/Linkurious/linkurious-server/blob/1b641a8c7a49c967dc0fe8dbc74faf5ed7793d9a/server/services/case/index.ts#L112

mzorkany (changes requested):

src/api/Alert/types.ts
EricMallard: it's also used in the frontend, that is why I created it here

mzorkany (approved):

PR #579, opened 2023-03-07
LKE-7492 Separate alert save from alert run

mzorkany:


src/api/Alert/types.ts
```diff
@@ -108,8 +105,19 @@ export interface IAlertQuery {
...
+type AlertError = {
+  queryId?: number;
+  source: 'caseAttributeQuery' | 'alertQuery';
+  error: LkError;
+  partial: boolean;
+};
+
 export interface IRunAlertParams extends IDataSourceParams {
   id: number;
+  waitForRun?: boolean;
```
davidrapin, line 117: still needed? have you check with SE?

davidrapin (commented):
first pass done

src/api/Alert/types.ts
mzorkany: Its still needed for the e2e tests as many of the tests use this API to run the alert and wait for the alert to run and the cases to be generated then check and validate the results

PR #590, opened 2023-03-17
LKE-8341 Add release lock API

Diachk:
fix: https://github.com/Linkurious/linkurious-server/pull/3902


src/http/request.ts
```diff
@@ -125,6 +131,20 @@ export abstract class Request<S = undefined> {
     };
   }
 
+  /**
+   * Send a post request using the Navigator.sendBeacon api.
+   * - Note that only url parameters are supported.
+   * - The sendBeacon api does not return any response.
+   */
+  public sendBeacon<EK extends LkErrorKey = never>(rawFetchConfig: SendBeaconConfig<EK>): void {
```
gmasclet, line 139: Consider Returning a promise, so that the return value of `releaseEditLock()` is always a promise:

```suggestion
  public async sendBeacon<EK extends LkErrorKey = never>(rawFetchConfig: SendBeaconConfig<EK>): Promise<void> {
```

gmasclet (approved):
Nice you've found the `navigator.sendBeacon()`, that's a great fit for our use case.

I note the frontend must register an event listener on `visibilitychange` to fully leverage this (releasing the lock when the user navigate to another view is not enough).

I've got one proposal to improve this pull request, see my comment. No need for a re-review, I trust you to do right changes.

src/http/request.ts
Diachk: Makes sense 👍

src/api/Visualization/index.ts
```diff
@@ -76,6 +77,29 @@ export class VisualizationAPI extends Request {
     });
   }
 
+  /**
+   * Release the exclusive edit lock on a visualization.
+   */
+  public releaseEditLock(params: ReleaseVisualizationEditLockParams) {
```
gmasclet, line 83: @hsenbassam This method is async, it returns a promise. As outlined by @Diachk, the call to `sendBeacon` has most likely to be done synchronously to guarantee that the request is scheduled, so it should be in an async function.

Let's not merge this pull request too quickly. We should first implements the frontend logic and carefully test that viz locks are released even if ones' browser is abruptly closed.

ping @aminerhanemi fyi

hsenbassam: Thank you @gmasclet for the info, I will based my frontend based on this rest-client and the one related in the server, we will not merge anything until we make sure everything is okay.

hsenbassam:
> visibilitychange

Actually I am not sure about [visibilitychange](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event) as I think this event is triggered also when you switch tab, which is not our case here, that's wy I relied on beforeunload event and onDestroy lifecycle hook.

Anyway we will give it a full test during the validation.

PR #591, opened 2023-03-20
LKE-8363 feat(UCL): hide modified queries

EricMallard:


src/api/Alert/types.ts
```diff
@@ -323,8 +322,10 @@ export interface ICaseColumn {
 
 export interface AlertQueryData {
   id: number;
+  modelKey: string;
```
davidrapin, line 325: Now that we have the modelKey, I don't think we always need the model ID.

There are probably different needs:
- listing alerts: we don't need the model ID but only the model key
- looking at a case in the case-list: same as above? (unsure)
- looking at the details of a case: we need the details of which exact versions of models matched

Taking that into account, maybe we can have two different objects:
- AlertQuery (no "ID" field)
-  AlertQueryVersion (with "ID" field)
maybe this distinction could help us make the server code simpler, since in many cases we need less details

davidrapin (commented):
I added a remark about always returning the model ID, let me know what you think.

src/api/Alert/types.ts
EricMallard: today we do not have a specific case where we need this id.  We have 2 types AlertQueryData and IAlertQuery that extends the first one. Knowing which of those could potentially require the id is not obvious today. So I would propose to remove it completely for now and see in the future with more context what types are needed.

EricMallard: actually I created this ALertQueryVersion because it's already needed in the server for deduplication.

EricMallard: The conclusion is yes, after digging into the code, we need the id. It is used in multiple places (update query, deduplicate query, etc), both front and back. We could migrate everything to model key, but:
- it's more complex (a lot to change in a lot of place, beginning by the interfaces)
- it's not obvious it's better, because we could argue that model key should only be used as a link between queries. For the rest, let's keep using the id.

EricMallard: eg: for deduplicating query, we can use modelKey today instead of the id. But the id behind it is to avoid having multiple matches with the same query, making us showing multiple time the same query in the frontend. 
In the future, if we need to show the exact version of a query that generated a case, we might re-use that code (with the id, not the model key), to delete duplicates and show that specific version in the frontend. Sorry, if it's not clear, we can have a quick chat about it.

davidrapin (approved):
Looking good!

PR #592, opened 2023-03-22
LKE-8218 feat(spaces): add missing return errors

aminerhanemi:


src/api/Visualization/index.ts
```diff
@@ -150,7 +151,7 @@ export class VisualizationAPI extends Request {
     params: ICreateVisualizationFolderParams
   ) {
     return this.request({
-      errors: [UNAUTHORIZED, DATA_SOURCE_UNAVAILABLE, ALREADY_EXISTS],
+      errors: [UNAUTHORIZED, DATA_SOURCE_UNAVAILABLE, ALREADY_EXISTS, NOT_FOUND, INVALID_PARAMETER],
```
gmasclet, line 154: I'm unsure about adding the `INVALID_PARAMETER` error key here.

According to the definition of the error keys, this key is part of the technical errors that should not be returned by the REST client, unless there is a bug. This makes sense as the REST client is not expected to send malformed requests to the server.

https://github.com/Linkurious/linkurious-rest-client/blob/6e5c99de7df6f5342cf78b23f471c52b89797bde/src/http/response.ts#L48-L51

I see from server E2E tests that this error key is returned if a folder cannot be created because its parent folder is not in the same space. We whould consider introducing a new `INVALID_PARENT_FOLDER` error key, so that the frontend is able to correctly differentiate and handle this error from an `INVALID_PARAMETER` response due to a REST client bug.

Let me know what you think.

aminerhanemi: Make sense to me. 
I will create a new error type and update the server accordingly

gmasclet (approved):
LGTM

PR #597, opened 2023-03-29
LKE-8423 Add last editing user in dashboard

Diachk:
fix: https://linkurious.atlassian.net/browse/LKE-8423


Diachk:
Closing, not need after all

PR #604, opened 2023-05-17
LKE-8650 Create a read comments on visualizations

mzorkany:


src/api/Visualization/types.ts
```diff
@@ -398,3 +398,26 @@ export type IUpdateWidgetParams = ICreateWidgetParams;
...
+  createdAt: string;
+}
+
+export interface CreateVisualizationCommentParams {
+  visualizationId: number;
+  content: string;
+}
+
+export interface GetVisualizationCommentParams {
+  // Extends Datasource params
```
hsenbassam, line 419: I think it should extends IDataSourceParams since we are passing the sourceKey in the path.

WDYT @Diachk ?

Diachk: Yes indeed it should !

Diachk (approved):
LGTM

src/api/Visualization/types.ts
```diff
@@ -398,3 +398,30 @@ export type IUpdateWidgetParams = ICreateWidgetParams;
...
+  content: string;
+}
+
+export interface GetVisualizationCommentsParams extends IDataSourceParams {
+  visualizationId: number;
+  offset?: number;
+  limit?: number;
+}
+
+export interface GetVisualizationCommentsResponse {
```
Diachk, line 424: Please use `PaginatedResponse<T>` instead for consistency

Diachk (changes requested):

Diachk (changes requested):

src/api/Visualization/types.ts
```diff
@@ -411,12 +411,12 @@ export interface VisualizationComment {
...
 
 export interface CreateVisualizationCommentParams extends IDataSourceParams {
-  visualizationId: number;
+  id: number;
   content: string;
 }
 
 export interface GetVisualizationCommentsParams extends IDataSourceParams {
-  visualizationId: number;
+  id: number;
```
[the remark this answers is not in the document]
mzorkany: @Diachk your suspicion is right, i modified the routes to use `:visualizationId` insted of `:id`

src/api/Visualization/types.ts
```diff
@@ -398,3 +398,30 @@ export type IUpdateWidgetParams = ICreateWidgetParams;
...
+  user: VisualizationCommentUser;
+  createdAt: string;
+}
+
+export interface CreateVisualizationCommentParams extends IDataSourceParams {
+  visualizationId: number;
+  content: string;
+}
+
+export interface GetVisualizationCommentsParams extends IDataSourceParams {
```
gmasclet, line 418: You can extend `PaginationClause`, so that  `offset` / `limit` don't have to be repeated:

```typescript
export interface GetVisualizationCommentsParams extends IDataSourceParams, PaginationClause {
  visualizationId: number;
}
```

Diachk (approved):
LGTM, removed an unused interface

PR #605, opened 2023-05-23
LKE-8759 feat(config): add localAuth flag

EricMallard:


src/api/Config/types.ts
```diff
@@ -252,6 +252,7 @@ export interface ILeafletConfig {
 export interface IAccessConfig {
   floatingLicenses?: number;
   authRequired?: boolean;
+  localAuth: boolean;
```
Diachk, line 255: Same comment as the server PR.
I have a preference for negative by default options it is more convenient for migrations.

```suggestion
  disableLocalAuth?: boolean;
```

Diachk (changes requested):
Review done, same comment as the server PR

Diachk (changes requested):
Review done, same comment as the server PR

Diachk (approved):
LGTM

PR #607, opened 2023-05-29
LKE-8651 feat(visualization): Delete visualization comment

aminerhanemi:


src/api/User/types.ts
```diff
@@ -8,6 +8,8 @@ import {GenericObject, IDataSourceParams, PersistedItem} from '../commonTypes';
 import {AccessRight, Action, AnyAction} from '../AccessRight';
 import {IGuestPreferencesConfig} from '../Config';
 
+export type DELETE_USER = 'Deleted user';
```
mzorkany, line 11: Since this is a type, let's follow the convention and rename the variable to `DeletedUser`

src/api/Visualization/index.ts
```diff
@@ -400,4 +402,16 @@ export class VisualizationAPI extends Request {
       params: params
     });
   }
+
+  /**
+   * Delete visualization comment
+   */
+  deleteComment(params: DeleteVisualizationCommentParams) {
```
mzorkany, line 409: Since the APIs for creating and getting comments are named 
`createVisualizationComment`
`getVisualizationComments`

lets follow the same pattern and rename this function to
`deleteVisualizationComment`

src/api/Visualization/types.ts
```diff
@@ -424,3 +424,8 @@ export interface CreateVisualizationCommentParams extends IDataSourceParams {
 export interface GetVisualizationCommentsParams extends IDataSourceParams, PaginationClause {
   visualizationId: number;
 }
+
+export interface DeleteVisualizationCommentParams extends IDataSourceParams {
+  commentId: number;
+  visualizationId: number;
```
mzorkany, line 430: Unless there is something that i am missing, i don't think the `visualizationId` is necessary since the `commentId` will always be unique

mzorkany (changes requested):
All good, just a few small changes

src/api/Visualization/types.ts
aminerhanemi: yeah it is not needed as the comment instance contain the id of the viz to which it belongs we can just getting from it.

mzorkany (approved):
All Good!

PR #616, opened 2023-06-27
LKE-9170 feat(spaces): Add getGroupUsers API

aminerhanemi:


src/api/User/index.ts
```diff
@@ -238,4 +239,16 @@ export class UserAPI extends Request {
       params: params
     });
   }
+
+  /**
+   * Get users by group id and return username, email and id
+   */
+  public getGroupUsers(this: Request<SearchUsersSimpleResponse>, params: GetGroupUsersParams) {
+    return this.request({
+      errors: [UNAUTHORIZED, FORBIDDEN, DATA_SOURCE_UNAVAILABLE, INVALID_PARAMETER],
```
gmasclet, line 248: From what I understand, we expect an `INVALID_PARAMETER` error if the group isn't found. I would rather use a `NOT_FOUND` error in this case, and keep `INVALID_PARAMETER` errors for actual validation issues (for instance a missing parameter).

```suggestion
      errors: [UNAUTHORIZED, FORBIDDEN, DATA_SOURCE_UNAVAILABLE, NOT_FOUND],
```

gmasclet (approved):
LGTM :+1:

I've just got a couple minor comments. No need for a re-review so I approve right away.

PR #617, opened 2023-07-03
LKE-9092 sort cases by model count

mzorkany:


EricMallard (approved):

src/api/Alert/types.ts
```diff
@@ -346,7 +346,8 @@ export enum FullCaseListSortProperties {
   STATUS = 'status',
   STATUS_CHANGED_BY = 'statusChangedBy',
   STATUS_CHANGED_ON = 'statusChangedOn',
-  ASSIGNEE = 'assignedUser'
+  ASSIGNEE = 'assignedUser',
+  ALERT_QUERIES_COUNT = 'alertQueriesCount'
```
davidrapin, line 350: why not call this ALERT_MODEL_COUNT? (I don't remember how we call it in other places in the code, but favor consistency)

mzorkany: @davidrapin 
The two terms are being used interchangeably, In the back-end side we always try to use the term `alertQuery` starting from the name of the DB table then all the way to functions and variable names

davidrapin (approved):

PR #622, opened 2023-07-25
LKE-9265: Adapt Alert column to support currency type

hsenbassam:


hsenbassam:
> @Diachk @aminerhanemi do you agree that reusing `ICurrencyOptions` is the way to go?

I am more convinced on your suggestion here: https://github.com/Linkurious/linkurious-server/pull/4101#discussion_r1279388694
So i will fix it

Diachk:
@hsenbassam @gmasclet  Makes sense to me to reuse `ICurrencyOptions` we have the exact same params and behaviour, only the scope is different.

hsenbassam:
> @hsenbassam @gmasclet Makes sense to me to reuse `ICurrencyOptions` we have the exact same params and behaviour, only the scope is different.



> @hsenbassam @gmasclet Makes sense to me to reuse `ICurrencyOptions` we have the exact same params and behaviour, only the scope is different.

@Diachk So I reused it but I moved it to the commonTypes as we have in the server and it is not only limited to GraphSchema

src/api/Alert/types.ts
```diff
@@ -31,8 +32,9 @@ export enum AlertColumnType {
 
 export interface IAlertColumn {
   type: AlertColumnType;
-  columnName?: string;
   columnTitle: string;
+  columnName?: string;
+  options?: ICurrencyOptions;
```
mzorkany, line 37: since the options property can be only undefined or ICurrencyOptions, lets rename this property to something more explicit like CurrencyOptions

src/api/Alert/types.ts
```diff
@@ -326,6 +328,7 @@ export interface ICaseColumn {
   type: AlertColumnType;
   columnValue: string | number | null;
   columnTitle: string;
+  options?: ICurrencyOptions;
```
mzorkany, line 331: same comment as above

Diachk (approved):
LGTM

PR #624, opened 2023-08-09
LKE-9429 filter cases by alertQueries model keys

mzorkany:


src/api/Alert/types.ts
```diff
@@ -102,6 +102,7 @@ export interface IAlertQuery extends AlertQueryData {
   query: string;
   dialect: GraphQueryDialect;
   updatedAt: Date;
+  alertId: number;
```
EricMallard, line 105: why adding this? the query is normally already inside the Alert type, so the alertId should be accessible

Diachk: Do we need to modify the api response for this reason ? Since the alert object contains alert queries.
I see that you haven't modified the API doc to reflect that.

Diachk (changes requested):
Review done, not sure I fully understand the need to include the alert id as well

Diachk (approved):
LGTM

PR #642, opened 2023-11-02
LKE-9867: Remove dev dependency chai

gmasclet:
https://linkurious.atlassian.net/browse/LKE-9867


davidrapin (approved):
Less dependencies <3

PR #643, opened 2023-11-02
LKE-9868: Update superagent to 8.1.2

gmasclet:
https://linkurious.atlassian.net/browse/LKE-9868

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,13 +9,13 @@
       "version": "4.0.17",
       "license": "MIT",
       "dependencies": {
-        "superagent": "6.1.0"
+        "superagent": "8.1.2"
       },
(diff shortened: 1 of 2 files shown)


davidrapin (approved):
@gmasclet works for me!

As a side note, if we don't align superagent to the same version (types and code) in rest-client, this will break umbrella.

davidrapin:
@gmasclet this seems to duplicate the changes from this other PR merged yesterday: https://github.com/Linkurious/linkurious-rest-client/pull/639/files

gmasclet:
@davidrapin Indeed, looks like we had a race condition here :smile:

I close this PR as #639 is already merged.

PR #644, opened 2023-11-03
LKE-9831 Mention users in case comments

mzorkany:


src/api/commonTypes.ts
```diff
@@ -92,3 +93,9 @@ export interface ICurrencyOptions {
   format: CurrencyFormat;
   symbol?: string;
 }
+
+export interface CommentMention {
+  position: number;
+  length: number;
+  user: MentionedUser;
+}
```
gmasclet, line 101: I suggest declaring this interface in '/src/api/User/types' (or moving `MentionedUser` and the related interfaces here), so that we avoid having a dependency cycle.

gmasclet (commented):
I've just got one minor comment, otherwise LGTM :+1:

mzorkany:
@gmasclet post review done :)

gmasclet (approved):
LGTM

PR #655, opened 2023-12-13
LKE-10018 refactor the widget service

mzorkany:


gmasclet (approved):
LGTM :+1:

As we complete the types here, I recommend simplifying the corresponding types in `linkurious-frontend` and `ogma-linkurious-parser`. From what I see at a quick glance:

* `WidgetContent` can be removed from https://github.com/Linkurious/linkurious-frontend/blob/1f360fc5a153906446f9744b17a522fcae9ea8e3/src/app/states/visualization/types/widget.ts#L25-L31

* `Legend` can be removed from https://github.com/Linkurious/ogma-linkurious-parser/blob/fa6fb724ca93c70e3cd78cc3edce743ff54f03f6/src/styles/styleRules.ts#L31-L33

PR #663, opened 2024-02-05
LKE-10301: Add the webhooks API

gmasclet:
https://linkurious.atlassian.net/browse/LKE-10301


src/api/Config/types.ts
```diff
@@ -132,6 +133,12 @@ export interface IGuestPreferencesConfig {
   uiFilter: boolean;
 }
 
+export interface WebhooksConfig {
+  deliveryFrequency?: string;
+  cleanupFrequency?: string;
+  deliveryRetentionDelayMs?: number;
```
Diachk, line 139: Isn't `cleanupFrequency` and `deliveryRetentionDelayMs` the same thing ?

src/api/webhook/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  events?: WebhookEvent[];
+}
+
+export interface DeleteWebhookParams {
+  webhookId: number;
+}
+
+export type PingWebhookParams = DeleteWebhookParams;
+
+export interface GetWebhookDeliveriesParams extends DeleteWebhookParams, PaginationClause {}
```
Diachk, line 21: The fact that multiple interfaces that have nothing to with delete extend the delete params looks weird in my opinion, it makes sense programatically but does not make sense functionally.

I think the cost to make functionally coherent is pretty low.
e.g:
```ts
interface ExistingWebhookParams {
  webhookId: number;
}

export interface DeleteWebhookParams extends ExistingWebhookParams {}

export interface PingWebhookParams extends ExistingWebhookParams {}

export interface GetWebhookDeliveriesParams extends ExistingWebhookParams, PaginationClause {}

```

Bonus point: Adding additional params to `DeleteWebhookParams` will not affect any other interface.

Diachk (changes requested):
Looks good to me overall, i just have a couple of remarks worth analysing.

src/api/webhook/types.ts
gmasclet: Good idea, it's done :+1:

src/api/Config/types.ts
gmasclet: * `cleanupFrequency` is a cron expression that controls the schedule of the cleanup task. 
* `deliveryRetentionDelayMs` is the amount of milliseconds after which a delivery becomes eligible for cleanup.

So there not the same thing. For instance, we may schedule the cleanup task once per day at midnight (`0 0 * * *`) and keep completed deliveries for one week (`1000 * 60 * 60 * 24 * 7 => 604800000`).

Diachk (approved):

PR #664, opened 2024-02-06
LKE-10215 Add createNodeGrouping rule route

mzorkany:


src/api/nodeGrouping/types.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright Linkurious SAS 2012 - 2024
+ *
+ * - Created on 2024-02-06.
+ */
+
+export interface NodeGroupingRule {
+  id: number;
+  sourceKey: string;
+  name: string;
+  itemType: string;
```
davidrapin, line 11: `itemType` and `propertyKey` are very specific to how the first mvp of node grouping works, but will not adapt well to future improvement to node-grouping. Can you suggest a more generic way to represent the grouping rule that will allow for future interations that will involve different kinds of grouping rules?

davidrapin (changes requested):
The NodeGroupingRule object is not future-proof, could you point me to the technical specs you wrote for this object? this design seems hard to evolve into future version of the feature with other kinds of node grouping rules.

src/api/nodeGrouping/types.ts
```diff
@@ -0,0 +1,34 @@
...
+export type NodeGroupingOptions = {
+  [NodeGroupingType.PROPERTY_KEY]: PropertyKeyNodeGroupingOptions;
+};
+
+export interface PropertyKeyNodeGroupingOptions {
+  itemType: string;
+  propertyKey: string;
+}
+
+export interface CreateNodeGroupingRuleParams extends Pick<NodeGroupingRule, 'sourceKey' | 'name'> {
```
aminerhanemi, line 29: @mzorkany I think we should extend IDataSourceParams where sourceKey is optional, which allows the frontend to call that API without sending each time the source key.
 The server gets the sourcekey value from the url

src/api/nodeGrouping/types.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Copyright Linkurious SAS 2012 - 2024
+ *
+ * - Created on 2024-02-06.
+ */
+
+export interface NodeGroupingRule {
+  id: number;
+  sourceKey: string;
```
aminerhanemi, line 9: instead of having sourcekey here, need to extend IDataSourceParams

mzorkany: @aminerhanemi extending the IDataSourceParams here does not make sense since the `sourceKey` is always defined on the `NodeGroupingRule` interface

i will modify `CreateNodeGroupingRuleParams` interface to extend `IDataSourceParams`

PR #667, opened 2024-02-08
LKE-10245 User wants to delete node group rule

mzorkany:


davidrapin (approved):
looking good!

src/api/nodeGrouping/types.ts
```diff
@@ -34,3 +34,5 @@ export interface CreateNodeGroupingRuleParams extends Pick<NodeGroupingRule, 'so
 }
 
 export interface GetNodeGroupingRulesParams extends Pick<NodeGroupingRule, 'sourceKey'> {}
+
+export interface DeleteNodeGroupingRuleParams extends Pick<NodeGroupingRule, 'id' | 'sourceKey'> {}
```
aminerhanemi, line 38: should extends IDataSourceParams

aminerhanemi (changes requested):

src/api/nodeGrouping/types.ts
mzorkany: done

PR #686, opened 2024-03-20
LKE-10703 feat(ownership): manage queries

EricMallard:


src/api/AccessRight/types.ts
```diff
@@ -22,6 +22,7 @@ export enum Action {
   RUN_QUERY = 'runQuery', // Execute a saved query
   RAW_READ_QUERY = 'rawReadQuery', // Create a read query
   RAW_WRITE_QUERY = 'rawWriteQuery', // Create a read/write query
+  RAW_MANAGE_QUERY = 'rawManageQuery', // Manage read/write queries
```
gmasclet, line 25: I don't think the `raw` prefix is needed here. Also worth enriching a bit the documentation:

```suggestion
  MANAGE_QUERY = 'manageQuery', // Manage all non-private read/write queries
```

EricMallard: > I don't think the `raw` prefix is needed here. Also worth enriching a bit the documentation:

it's a bit late and would require multiple changes in multiple repos when we are about to merge. I can open a ticket and do it next sprint...good for you?

PR #687, opened 2024-03-20
LKE-10712 Manage custom action access right

mzorkany:


src/api/AccessRight/types.ts
```diff
@@ -23,7 +23,8 @@ export enum Action {
   RAW_READ_QUERY = 'rawReadQuery', // Create a read query
   RAW_WRITE_QUERY = 'rawWriteQuery', // Create a read/write query
   RUN_CUSTOM_ACTION = 'runCustomAction', // Execute a custom action
-  WRITE_CUSTOM_ACTION = 'writeCustomAction' // Edit, update and delete a custom action
+  WRITE_CUSTOM_ACTION = 'writeCustomAction', // Edit, update and delete a custom action
+  MANAGE_CUSTOM_ACTION = 'manageCustomAction' // Create, edit, update and delete a custom action
```
gmasclet, line 27: It's worth documenting more accurately the difference between these two rights here. Otherwise, this looks confusing someone without context.

From my understanding:
* `WRITE_CUSTOM_ACTION`: create a custom action + update/delete owned custom actions
* `MANAGE_CUSTOM_ACTION`: `WRITE_CUSTOM_ACTION` + update/delete all non-private custom actions

src/api/CustomAction/types.ts
```diff
@@ -61,7 +61,7 @@ export interface ParsedCustomAction {
 }
 
 export enum CustomActionRight {
-  OWNER = 'owner',
+  MANAGE = 'manage',
```
gmasclet, line 64: `write` seems more idiomatic to me, especially in conjunction with `read`.

It's also more accurate in my opinion as:
* if I'm a query manager, I've got `write` access on all the non-private queries.
* if I'm the owner of a query, I've got `write` access on that query, but I don't have the right to `manage` other queries.

```suggestion
  WRITE = 'write',
```

gmasclet: ping @EricMallard as this is also applicable to https://github.com/Linkurious/linkurious-rest-client/pull/686

mzorkany: @gmasclet @EricMallard 
I already answered in the slack discussion :)

mzorkany:
@gmasclet Post review done :)

src/api/CustomAction/types.ts
gmasclet: As agreed on slack, let's stick with 'manage'. I close this comment.

gmasclet (approved):
LGTM :+1:

PR #717, opened 2024-05-23
LKE-10684 make LKErrorKey enum comprehensive

mzorkany:


src/http/response.ts
```diff
@@ -19,11 +19,13 @@ export enum LkErrorKey {
   // Business and access errors
   FEATURE_DISABLED = 'feature_disabled',
   UNAUTHORIZED = 'unauthorized',
+  LDAP_BIND_ERROR = 'ldap_bind_error',
```
gmasclet, line 22: This error key is actually no longer used. See https://github.com/Linkurious/linkurious-server/pull/4649#discussion_r1581249386

So we don't need to add this entry.

src/http/response.ts
```diff
@@ -19,11 +19,13 @@ export enum LkErrorKey {
...
   DATA_SOURCE_UNAVAILABLE = 'dataSource_unavailable',
   GUEST_DISABLED = 'guest_disabled',
   FORBIDDEN = 'forbidden',
   NOT_FOUND = 'not_found',
   BAD_GRAPH_REQUEST = 'bad_graph_request',
+  CREATION_FAILED = 'creation_failed',
```
gmasclet, line 28: Unless I'm wrong, this error key is not used, so there is no need to add it here.

src/http/response.ts
```diff
@@ -50,6 +53,9 @@ export enum LkErrorKey {
   CRITICAL = 'critical',
   BUG = 'bug',
   SOCKET_ERROR = 'socket_error',
+  PORT_BUSY = 'port_busy',
+  PORT_RESTRICTED = 'port_restricted',
```
gmasclet, line 57: Looks like these two error keys are currently unused. However, they should be used in https://github.com/Linkurious/linkurious-server/blob/493cf372002e28a13580298d58894eb299b693cb/server/services/webServer/index.ts#L391-L410, so let's add them here.

gmasclet (approved):

PR #718, opened 2024-05-28
LKE-11084 User can edit a node grouping rule

mzorkany:


src/api/nodeGrouping/types.ts
```diff
@@ -35,6 +40,12 @@ export interface CreateNodeGroupingRuleParams extends IDataSourceParams {
   groupingOptions: NodeGroupingOptions[NodeGroupingType];
 }
 
+export interface UpdateNodeGroupingRuleParams
+  extends Omit<CreateNodeGroupingRuleParams, 'uuid' | 'name'> {
```
Diachk, line 44: There is a mismatch between the type declaration and the validation this can cause some bugs.
```suggestion
  extends Partial<Omit<CreateNodeGroupingRuleParams, 'uuid' | 'name'>> {
```

Diachk (commented):
Review done, added to comment to mirror the review on the server side

PR #739, opened 2024-08-22
LKE-11806 QueryTimeout DB config

mzorkany:
LKE-11806: Add the queryTimeout option as a database config. 
This will enable the users to configure the timeout limit for the database queries


src/api/Config/types.ts
```diff
@@ -69,6 +68,7 @@ export interface IDatabaseConfig {
   username?: string;
   password?: string;
   connectionRetries?: number;
+  queryTimeout?: number;
```
davidrapin, line 71: I suggest "transactionTimeout", I reflects the reality and is standard in the DB world. cc @Diachk 


```suggestion
  transactionTimeout?: number;
```

davidrapin (changes requested):

davidrapin (approved):

PR #740, opened 2024-08-22
LKE-7018 Rename indexation to indexing

Diachk:
jira: LKE-7018

This renames indexation to indexing wherever visible to a user.


davidrapin:
@enys fyi, if you're using any of these configuration keys, you might need to update the keys.

@Diachk i initially thought the update was only about labels in the frontend. I'm okay with updating config keys this too, but there is a little bit more risk of breaking existing automation/scripts.

davidrapin (approved):

PR #742, opened 2024-08-27
LKE-11783 feat(TagsManagement): get tags with associated queries

aminerhanemi:


src/api/GraphQuery/types.ts
```diff
@@ -225,7 +225,7 @@ export interface IGetQueryParams extends IDataSourceParams {
 }
 
 export interface IGetQueriesParams extends IDataSourceParams {
-  type: GraphQueryType;
+  type?: GraphQueryType;
```
aminerhanemi, line 228: In the current implementation to get all queries we need to call the API twice, once with `GraphQueryType.STATIC` and another time with `GraphQueryType.TEMPLATE`. Then we filter the results in [getQueries](https://github.com/Linkurious/linkurious-server/blob/1f35ac8017da9b2700455001b4075c3a44f22adb/server/services/graphQuery/index.ts#L334)

In the current state, the frontend always retrieves the full list of queries. Making  type optional allows us to make only one request

Diachk (approved):
LGTM

hsenbassam (approved):

PR #758, opened 2024-10-29
LKE-12088 data model for query builder

davidrapin:


mzorkany (approved):
Good for me

I like the idea of separating the types by functionality instead of dumping everything in one file.

cc @gmasclet 
I think we should do the same for the other resources, at least the ones that have chunky `types` file like alerts for example

T1B0 (approved):

davidrapin:
@aminerhanemi oh good point, let's change that, it's not a must have

davidrapin:
@aminerhanemi done via https://github.com/Linkurious/linkurious-rest-client/commit/e7e3647af50c29e74ba55e6495323c72ce117cf4 in the LKE-12219-query-builder-v1 branch

PR #778, opened 2024-11-26
LKE-12414 Graph query stats

mzorkany:


src/api/GraphQuery/queryCRUD.ts
```diff
@@ -102,3 +102,21 @@ export interface ConvertQueryResponse {
...
   dialect: GraphQueryDialect;
 }
+
+export interface GetQueryStatsParams extends IDataSourceParams {
+  query: StructuredGraphQuery;
+  dialect: GraphQueryDialect.QUERY_BUILDER;
+}
+
+export interface GetQueryStatsResponse {
+  resultCount: {
```
hsenbassam, line 112: is it better to have accuracy and value in some dedicated interface ?
```
 resultCount: QueryStatsResultCount

export interface QueryStatsResultCount {
    accuracy: QueryStatsCountAccuracy;
    value: number;
}
```

src/api/GraphQuery/queryCRUD.ts
```diff
@@ -102,3 +102,21 @@ export interface ConvertQueryResponse {
   query: string;
   dialect: GraphQueryDialect;
 }
+
+export interface GetQueryStatsParams extends IDataSourceParams {
+  query: StructuredGraphQuery;
+  dialect: GraphQueryDialect.QUERY_BUILDER;
```
hsenbassam, line 108: As dialect is always queryBuilder, so it is mandatory to always pass it as param from the request ?

src/api/GraphQuery/queryCRUD.ts
gmasclet: I'm not sure I see the benefit of introducing this interface. In any case, we can add it later if we change our minds.

src/api/GraphQuery/queryCRUD.ts
gmasclet: The idea of Mahmoud was to prepare the API to support other dialect. That's debatable as we could also have considered adding the field later.

PR #783, opened 2024-12-01
LKE-12483: Add NULL and NOT NULL operators

hsenbassam:
https://linkurious.atlassian.net/browse/LKE-12482

### DIFF
--- a/src/api/GraphQuery/queryBuilder.ts
+++ b/src/api/GraphQuery/queryBuilder.ts
@@ -164,7 +164,7 @@ export interface GraphQueryEdgeFilter {
  */
 export interface BasePropertyFilter {
   propertyType: 'string' | 'number' | 'date';
-  operator: '=' | '!=' | '<' | '<=' | '>' | '>=' | '~' | 'isNull' | 'isNotNull';
+  operator: BaseOperators;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/api/GraphQuery/queryBuilder.ts
```diff
@@ -211,12 +211,12 @@ export interface BaseStrictPropertyFilter<T> extends BasePropertyFilter {
 
 export interface StringStrictPropertyFilter extends BaseStrictPropertyFilter<string[]> {
   propertyType: 'string';
-  operator: '=' | '!=';
+  operator: '=' | '!=' | 'null' | 'notNull';
```
davidrapin, line 214: suggestion: using « isNull » and « isNotNull », because « null » looks like someone casted `null` into a string by mistake.

src/api/GraphQuery/queryBuilder.ts
```diff
@@ -211,12 +211,12 @@ export interface BaseStrictPropertyFilter<T> extends BasePropertyFilter {
...
 export interface StringStrictPropertyFilter extends BaseStrictPropertyFilter<string[]> {
   propertyType: 'string';
-  operator: '=' | '!=';
+  operator: '=' | '!=' | 'isNull' | 'isNotNull';
 }
 
 export interface NumberStrictPropertyFilter extends BaseStrictPropertyFilter<number> {
   propertyType: 'number';
-  operator: '>' | '<' | '=' | '<=' | '>=' | '!=';
+  operator: '>' | '<' | '=' | '<=' | '>=' | '!=' | 'isNull' | 'isNotNull';
```
gmasclet, line 219: ** Optional suggestion**

We could consider introducing a type for these operators, I think it would be useful in the backend code.

```typescript
// This is just a simplified example, we have actually number and string operators, and some of them are common
export const QUERY_BUILDER__OPERATORS = ['>' | '<' | '=' | '<=' | '>=' | '!=' | 'isNull' | 'isNotNull'] as const;
export type QueryBuilerOperator = (typeof OPERATORS)[number];
```

src/api/GraphQuery/queryBuilder.ts
```diff
@@ -208,15 +208,33 @@ export interface BaseStrictPropertyFilter<T> extends BasePropertyFilter {
...
+  'isNull',
+  'isNotNull'
+] as const;
+export type BaseOperators = (typeof BASE_OPERATORS)[number];
+
+export const STRING_OPERATORS = ['=', '!=', 'isNull', 'isNotNull'] as const;
+export type StringOperators = (typeof STRING_OPERATORS)[number];
+
+export const NUMBER_OPERATORS = ['>', '<', '=', '<=', '>=', '!=', 'isNull', 'isNotNull'] as const;
+export type NumberOperators = (typeof NUMBER_OPERATORS)[number];
```
gmasclet, line 228: We should use the singular form for the type:

```suggestion
export type BaseOperator = (typeof BASE_OPERATORS)[number];

export const STRING_OPERATORS = ['=', '!=', 'isNull', 'isNotNull'] as const;
export type StringOperator = (typeof STRING_OPERATORS)[number];

export const NUMBER_OPERATORS = ['>', '<', '=', '<=', '>=', '!=', 'isNull', 'isNotNull'] as const;
export type NumberOperator = (typeof NUMBER_OPERATORS)[number];
```

gmasclet (approved):
LGTM

PR #801, opened 2025-02-24
LKE-12859 spanner: adding a "var" field in graph-input templates 

davidrapin:
LKE-12859
to be able to set the var name to use for nodes/edges in the node/edge filter statement.


davidrapin:
nevermind, found a better way

PR #803, opened 2025-02-28
LKE-12832: Add the "explain why" entity resolution API

gmasclet:
https://linkurious.atlassian.net/browse/LKE-12832

This pull request adds some three resolution endpoints:
* explain why a record resolved to an entity.
* explain why two entities are related.
* retrieve the details of an entity
  * The list of its records ("Matched records").
  * The other entities that are possibly same ("Possible match" / "POSSIBLY_SAME") or possibly related ("Possible relationships" / "POSSIBLY_RELATED").

The "GET entity detail by entity Id" endpoint is one possible solution that basically does a "node expand" under the hood. 
* pros: it's straighforward and accurate. It includes results that may not be part of the visualization.
* cons: it has to be called by the frontend when the "Entity resolution" panel is displayed, which may induce some latency.

Otherwise, we might consider adding these entity resolution details in the "GET visualization" endpoint response and eagerly loading them for all the entities of the viz at once. I suspect this is a bad idea. In particular, it could be computationally taxing if the viz is large and/or contains many entities.

### DIFF
--- a/src/api/entityResolution/index.ts
+++ b/src/api/entityResolution/index.ts
@@ -16,7 +16,9 @@ import {
   EntityResolutionMetrics,
   ExplainWhyEntitiesParams,
   ExplainWhyRecordParams,
+  GetEntityByIdParams,
   IngestionStatus,
+  ResolvedEntity,
   StartEntityResolutionTaskParams,
   UpdateEntityResolutionMappingParams,
   WhyEntities,
@@ -229,6 +231,24 @@ export class EntityResolutionAPI extends Request {
         ENTITY_RESOLUTION_EXPIRED_LICENSE
       ],
       url: '/:sourceKey/entityResolution/why/entities/:entityId/:otherEntityId',
+      method: 'GET',
+      params: params
+    });
+  }
+
+  /**
+   * Explain why two different entities are related.
+   */
+  getEntityById(this: Request<ResolvedEntity>, params: GetEntityByIdParams) {
+    return this.request({
+      errors: [
+        UNAUTHORIZED,
+        FORBIDDEN,
+        DATA_SOURCE_UNAVAILABLE,
+        NOT_FOUND,
+        ENTITY_RESOLUTION_EXPIRED_LICENSE
+      ],
+      url: '/:sourceKey/entityResolution/entities/:entityId',
       method: 'GET',
       params: params
     });
--- a/src/api/entityResolution/index.ts
+++ b/src/api/entityResolution/index.ts
@@ -14,9 +14,13 @@ import {
   EntityResolutionLicenseInfo,
   EntityResolutionMapping,
   EntityResolutionMetrics,
+  ExplainWhyEntitiesParams,
+  ExplainWhyRecordParams,
   IngestionStatus,
   StartEntityResolutionTaskParams,
-  UpdateEntityResolutionMappingParams
+  UpdateEntityResolutionMappingParams,
+  WhyEntities,
+  WhyRecord
 } from './types';
 
 export * from './types';
@@ -189,6 +193,42 @@ export class EntityResolutionAPI extends Request {
     return this.request({
       errors: [UNAUTHORIZED, FORBIDDEN, DATA_SOURCE_UNAVAILABLE],
       url: '/:sourceKey/entityResolution/metrics',
+      method: 'GET',
+      params: params
+    });
+  }
+
+  /**
+   * Explain why a given record resolved to an entity.
+   */
+  explainWhyRecord(this: Request<WhyRecord>, params: ExplainWhyRecordParams) {
+    return this.request({
+      errors: [
+        UNAUTHORIZED,
+        FORBIDDEN,
+        DATA_SOURCE_UNAVAILABLE,
+        NOT_FOUND,
+        ENTITY_RESOLUTION_EXPIRED_LICENSE
+      ],
+      url: '/:sourceKey/entityResolution/why/record/:recordId',
+      method: 'GET',
+      params: params
+    });
+  }
+
+  /**
+   * Explain why two different entities are related.
+   */
+  explainWhyEntities(this: Request<WhyEntities>, params: ExplainWhyEntitiesParams) {
+    return this.request({
+      errors: [
+        UNAUTHORIZED,
+        FORBIDDEN,
+        DATA_SOURCE_UNAVAILABLE,
+        NOT_FOUND,
+        ENTITY_RESOLUTION_EXPIRED_LICENSE
+      ],
+      url: '/:sourceKey/entityResolution/why/entities/:entityId/:otherEntityId',
       method: 'GET',
       params: params
     });
--- a/src/api/entityResolution/index.ts
+++ b/src/api/entityResolution/index.ts
@@ -237,7 +237,7 @@ export class EntityResolutionAPI extends Request {
   }
 
   /**
-   * Explain why two different entities are related.
+   * Get a resolved entity by its ID
    */
   getEntityById(this: Request<ResolvedEntity>, params: GetEntityByIdParams) {
     return this.request({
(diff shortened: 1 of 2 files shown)


WilliamGorge:
@gmasclet @aminerhanemi i added https://github.com/Linkurious/linkurious-rest-client/pull/803/commits/284b5354b2fe582c1e086b3550e9786cd59df358 because i don't want to fall into premature optimisation and i don't think this call is costly.

Let's see if we have feedback, and maybe we can think of optimisation strategies such as doing some prefetch frontend-wise when selecting the done for example

src/api/entityResolution/types.ts
```diff
@@ -395,3 +395,144 @@ export interface EntityResolutionMetrics {
...
+   * The ID of the other related entity.
+   */
+  otherEntityId: number;
+}
+
+export interface GetEntityByIdParams extends IDataSourceParams {
+  /**
+   * The ID of the first entity to explain. It corresponds to the `entityId` property of the entity
+   * node in the graph database.
+   */
```
WilliamGorge, line 423: wrong copy paste and in the method

src/api/entityResolution/types.ts
```diff
@@ -395,3 +395,144 @@ export interface EntityResolutionMetrics {
...
+export interface RelatedEntity extends EntityResolutionMatch {
+  id: number;
+  name: string;
+  ambiguous: boolean;
+}
+
+export interface EntityResolutionMatch {
+  matchKey: MatchKey;
+  matchLevel: number;
+  matchLevelCode: `${MatchLevel}`;
```
gmasclet, line 528: No need to use a template literal type here if MatchLevel is an union type:

```suggestion
  matchLevelCode: MatchLevel;
```

src/api/entityResolution/types.ts
```diff
@@ -395,3 +395,144 @@ export interface EntityResolutionMetrics {
...
+  | 'Telegram'
+  | 'Twitter'
+  | 'Viber'
+  | 'Website'
+  | 'Wechat'
+  | 'WhatsApp'
+  | 'Zoom room';
+
+// Internal note: this type is also used by LKE's entity resolution client
+export interface ResolvedEntity extends IDataSourceParams {
```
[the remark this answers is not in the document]
gmasclet: * `IDataSourceParams` is meant for request params. This is a response, so `IDataSourceParams` should not be used here:

```typescript
export interface ResolvedEntity {
  sourceKey: string
  // ...
}
```

* We should consider exposing the regular source-key to the frontend. Looks fishy to me to expose the fully qualified source key here, we're going to jump into issues if one decides to use this source-key.

src/api/entityResolution/types.ts
```diff
@@ -395,3 +395,144 @@ export interface EntityResolutionMetrics {
...
+  entityResolutionRuleCode: string;
+}
+
+// Internal note: this type is also used by LKE's entity resolution client
+export enum MatchLevel {
+  RESOLVED = 'RESOLVED',
+  POSSIBLY_SAME = 'POSSIBLY_SAME',
+  POSSIBLY_RELATED = 'POSSIBLY_RELATED'
+  // These are known match level codes, some might be added later
+}
```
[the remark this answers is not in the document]
WilliamGorge: we already had this debate here with david https://github.com/Linkurious/linkurious-server/pull/5179#discussion_r1950003428

we want to avoid magic string usage in the backend, there is no better way to do that, we can have this discussion again if you want...

src/api/entityResolution/types.ts
```diff
@@ -395,3 +395,144 @@ export interface EntityResolutionMetrics {
...
+  | 'Tax ID'
+  | 'Telegram'
+  | 'Twitter'
+  | 'Viber'
+  | 'Website'
+  | 'Wechat'
+  | 'WhatsApp'
+  | 'Zoom room';
+
+// Internal note: this type is also used by LKE's entity resolution client
```
gmasclet, line 506: I'm not convinced these comments are useful, but I'm OK if you'd like to keep them.

`LKE's entity resolution client` seems confusing to me. I suggest calling it `entity resolution client` to makes it clearer we're not talking of the LKE REST Client (the one in this repo).

src/api/entityResolution/types.ts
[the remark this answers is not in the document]
gmasclet: Interestingly David made the exact same proposal as the one I'm doing here.

Without going into a lengthly discussion, I think union types are generally better than enum (mainly for the reasons enumerated in David's comment).

It's debatable whether strings are really magic ones as soon as they're typed with a union (they become checked at compile time).

I don't want to block this PR, you can move forward with an enum. Let's have a talk later.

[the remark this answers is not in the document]
WilliamGorge: > It's debatable whether strings are really magic ones as soon as they're typed with a union (they become checked at compile time).

it's not the case because you can't really static type a map where keys are actually node categories or edges types. Here the enum really has value for this particular case: you need to be able to use the type as a value

src/api/entityResolution/types.ts
```diff
@@ -395,3 +395,144 @@ export interface EntityResolutionMetrics {
...
+export interface ExplainWhyEntitiesParams extends IDataSourceParams {
+  /**
+   * The ID of the first entity to explain. It corresponds to the `entityId` property of the entity
+   * node in the graph database.
+   */
+  entityId: number;
+  /**
+   * The ID of the other related entity.
+   */
+  otherEntityId: number;
```
aminerhanemi, line 416: what is entityId?  is it the node id? if so it should be a string and not number similar to recordId

WilliamGorge: entityId is an id managed by senzing, you have it from the `ResolvedPerson/OrganisationEntity` node properties. It's a number because Senzing decided so.

Records are objects we send to senzing, so their id is what we decide, in our case it's the graph nodes and they are string because we decided so for LKE.

src/api/entityResolution/types.ts
WilliamGorge: https://github.com/Linkurious/linkurious-rest-client/pull/803/commits/a1485084cad78ca13e0a956de094b34e500a1237

src/api/entityResolution/types.ts
WilliamGorge: discussed in https://github.com/Linkurious/linkurious-rest-client/pull/803#discussion_r1981043107

src/api/entityResolution/index.ts
```diff
@@ -193,4 +199,58 @@ export class EntityResolutionAPI extends Request {
...
+        ENTITY_RESOLUTION_EXPIRED_LICENSE
+      ],
+      url: '/:sourceKey/entityResolution/why/entities/:entityId/:otherEntityId',
+      method: 'GET',
+      params: params
+    });
+  }
+
+  /**
+   * Explain why two different entities are related.
```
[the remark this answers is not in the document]
WilliamGorge: https://github.com/Linkurious/linkurious-rest-client/pull/803/commits/e8533defece1f63e82252e90f8ef88b5562c6543

WilliamGorge (approved):

src/api/entityResolution/types.ts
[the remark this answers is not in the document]
WilliamGorge: i fixed it here:
https://github.com/Linkurious/linkurious-rest-client/pull/803/commits/01d97f2e573bf96b4c6aa7a1cd56f34fdcfd03c4

and in the back; I removed the sourceKey from the response and j'en ai profité pour ajouter un check important je pense:
https://github.com/Linkurious/linkurious-server/pull/5212/files#diff-251bc0abda1e013db9f4ae80999d605812b795c6a3b857fbaa2159b4b0c0543e

gmasclet:
@WilliamGorge LGTM :+1:

src/api/entityResolution/types.ts
```diff
@@ -395,3 +395,142 @@ export interface EntityResolutionMetrics {
...
+  different: EntityAttributeKey[];
+  ambiguous: boolean;
+}
+
+interface MatchScore<T> {
+  key: EntityAttributeKey;
+  values: T[];
+}
+
+interface ScoredValue {
```
aminerhanemi, line 456: we need to export this interface to use it in the front

PR #804, opened 2025-03-10
LKE-12948 feat(ER) Handle feature disabled error response

hsenbassam:
Context: In the frontend, this error is not returned as a response so I can easily check if the error is feature_disabled (we used to check the error in the frontend in case we need to react based on it), as it is not part of the errors I need to try/catch in order to get it and then I cannot easily access the error key

PR: Add the FEATURE_DISABLED to the list of errors of ER API


src/api/entityResolution/index.ts
```diff
@@ -170,7 +171,13 @@ export class EntityResolutionAPI extends Request {
    */
   getIngestionStatus(this: Request<IngestionStatus>, params: IDataSourceParams) {
     return this.request({
-      errors: [UNAUTHORIZED, FORBIDDEN, DATA_SOURCE_UNAVAILABLE, ILLEGAL_SOURCE_STATE],
+      errors: [
+        UNAUTHORIZED,
+        FORBIDDEN,
+        DATA_SOURCE_UNAVAILABLE,
+        ILLEGAL_SOURCE_STATE,
+        FEATURE_DISABLED
```
hsenbassam, line 179: @gmasclet do we need the same in the why/why not APIs ?
Context: In the frontend, this error is not returned as a response so I can easily check if the error is feature_disabled (we used to check the error in the frontend in case we need to react based on it), as it is not part of the errors I need to try/catch in order to get it and then I cannot easily access the error key

gmasclet: @hsenbassam I agree with this change, it's consistent with what we're doing on other parts of the API (In particular, I've checked the alerts).

The `FEATURE_DISABLED` error may be returned by all the entity resolution endpoints. See https://github.com/Linkurious/linkurious-server/blob/521e29e0363b262e27dcb7081b45d7ff64a83dd5/server/routes/entityResolution.ts#L45-L56

So my recommendation is to declare this error on all the methods in this file.

hsenbassam:
Kept on Hold until we work on the story, so we can test everything properly

hsenbassam:
Closing as the main ticket will not be implemented, in case we want this improvement we can open a dedicated small ticket for it

PR #805, opened 2025-03-10
LKE-12943 feat(entity-resolution): group entities

EricMallard:


davidrapin (approved):

src/api/nodeGrouping/types.ts
```diff
@@ -14,25 +14,44 @@ export interface NodeGroupingRule {
...
+export type NodeGroupingRule = NodeGroupingByPropertyValue | NodeGroupingByAdjacentEdgeType;
 
 export enum NodeGroupingRuleRight {
   MANAGE = 'manage',
   READ = 'read'
 }
 
 export enum NodeGroupingType {
-  PROPERTY_KEY = 'propertyKey'
+  BY_PROPERTY_VALUE = 'propertyKey',
```
hsenbassam, line 35: why the value is "propertyKey" while the key is "property value"

davidrapin: @hsenbassam good question! worth a code comment I think:
- in practice, the feature is grouping nodes by property value (e.g. "nodes with property country has the same value")
- the enum was defined in a past version and migrating it is a pain now

EricMallard: explained

PR #809, opened 2025-03-17
LKE-13045 feat(NodeGrouping): add NodeGroupingConfig to global advanced config

aminerhanemi:


src/api/Config/types.ts
```diff
@@ -466,3 +467,7 @@ export interface EntityResolutionConfig {
   entityNodeCategories?: Partial<Record<EntityResolutionRecordType, string>>;
   serviceApiKey?: string;
 }
+
+export interface NodeGroupingConfig {
+  collapsed?: boolean;
```
EricMallard, line 472: why not put a default value at false in the backend? It would allow us to always have this value and avoid putting a default value in the frontend

EricMallard (approved):
Approving, small comment but not blocking

src/api/Config/types.ts
aminerhanemi: The default configuration is stored in config files. I already set the default value, but what if someone updates from an older version? or just remove the value from the config file? 
I think for some specific fields the server will throw an error if they are not defined but I don't think it should be the case here.
If it was in the SQL data based we could have defined a default value

EricMallard: yeah you're right, you'd need a migration, not worth it

PR #815, opened 2025-04-01
LKE-131013 feat(Active/Active): Add Cluster config types

hsenbassam:
**Jira Ticket:**

https://linkurious.atlassian.net/browse/LKE-13103

**Description:**

Add Cluster Config Types

### DIFF
--- a/src/api/Config/types.ts
+++ b/src/api/Config/types.ts
@@ -464,6 +464,11 @@ export interface IEmailNotificationsConfig {
   fromEmail: string;
 }
 
+export enum ClusterMode {
+  PRIMARY = 'primary',
+  SECONDARY = 'secondary'
+}
+
 export interface EntityResolutionConfig {
   enabled: boolean;
   url: string;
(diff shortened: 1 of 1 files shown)


src/api/Config/types.ts
```diff
@@ -462,11 +463,19 @@ export interface IEmailNotificationsConfig {
...
 export interface EntityResolutionConfig {
   enabled: boolean;
   url: string;
   chunkSize?: number;
   expandLimitPerNode?: number;
   serviceApiKey?: string;
 }
+
+export const CLUSTER_MODE = ['primary', 'secondary'] as const;
+export type ClusterMode = (typeof CLUSTER_MODE)[number];
```
mzorkany, line 475: We can define ClusterMode as type ClusterMode = 'primary' | 'secondary'  

More clear and we can remove the need for the const `CLUSTER_MODE`

mzorkany (commented):
Review done:
Only one small change

src/api/Config/types.ts
hsenbassam: I think in this case we cannot use   @values(CLUSTER_MODE), as clusterMode is type and not string[], we can use   @values(['primary', 'secondary'])

mzorkany (approved):

PR #817, opened 2025-04-03
LKE-11587  User Update: Add FEATURE_DISABLED to errors list

justinmarsan:
LKE-11587

When a user is from source SSO with `access.autoRefreshGroupMapping` settings, their groups can't be changed, and return a FEATURE_DISABLED type of error that was not part of the expected list, preventing the frontend from catching that error and displaying proper error messaging.

### DIFF
--- a/src/api/User/index.ts
+++ b/src/api/User/index.ts
@@ -35,6 +35,7 @@ import {
 export * from './types';
 
 const {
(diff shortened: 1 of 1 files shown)


WilliamGorge (approved):
thanks

PR #827, opened 2025-04-30
LKE-12450 Stop ER ingestion

mzorkany:
https://linkurious.atlassian.net/browse/LKE-12540


src/api/entityResolution/index.ts
```diff
@@ -148,6 +150,24 @@ export class EntityResolutionAPI extends Request {
...
+  /**
+   * Stop an ingestion task on a given data-source.
+   */
+  stopIngestion(this: Request<IngestionStatus>, params: StopEntityResolutionTaskParams) {
+    return this.request({
+      errors: [
+        UNAUTHORIZED,
+        FORBIDDEN,
+        DATA_SOURCE_UNAVAILABLE,
+        NOT_SUPPORTED,
```
davidrapin, line 162: when does this error happen?

src/api/entityResolution/index.ts
```diff
@@ -148,6 +150,24 @@ export class EntityResolutionAPI extends Request {
...
 
+  /**
+   * Stop an ingestion task on a given data-source.
+   */
+  stopIngestion(this: Request<IngestionStatus>, params: StopEntityResolutionTaskParams) {
+    return this.request({
+      errors: [
+        UNAUTHORIZED,
+        FORBIDDEN,
+        DATA_SOURCE_UNAVAILABLE,
```
davidrapin, line 161: if the source is offline, wouldn’t the ER be stopped by itself?

davidrapin (approved):
added 2 questions, not blocking

src/api/entityResolution/index.ts
mzorkany: This error is just a contingency in case the `stop` API is called on a datasource that does not support ER

src/api/entityResolution/index.ts
mzorkany: yes, but also the API can be called with a sourceKey of an offline datasource

PR #830, opened 2025-05-12
LKE-13159 Add information to LicenseInfo

justinmarsan:


src/api/Config/types.ts
```diff
@@ -22,6 +22,7 @@ export interface Configuration {
   ssoProvider?: 'oauth2' | 'saml2';
   url: string;
   setupAuthentication: boolean;
+  instanceKey?: string;
```
mzorkany, line 25: why optional, this key is always defined and its always returned as part of the config response
https://github.com/Linkurious/linkurious-server/blob/69a6aad0aa4a17473e31809ea2f75b4e84e27aab/server/models/config/LKEConfig.ts#L115

PR #831, opened 2025-05-19
LKE-13142  feat(ER): add API to get ER server status

aminerhanemi:


src/api/entityResolution/types.ts
```diff
@@ -570,3 +570,7 @@ export enum MatchLevel {
   POSSIBLY_RELATED = 'POSSIBLY_RELATED'
   // These are known match level codes, some might be added later
 }
+
+export interface GetEntityResolutionServerStatusParams extends IDataSourceParams {}
+
+export type EntityResolutionServerStatus = 'up' | 'down';
```
EricMallard, line 576: why not a boolean?

EricMallard (approved):
optional comment

src/api/entityResolution/index.ts
```diff
@@ -273,4 +274,25 @@ export class EntityResolutionAPI extends Request {
       params: params
     });
   }
+
+  /**
+   * Get entity resolution server status
+   */
+  getEntityResolutionServerStatus(
+    this: Request<{status: string}>,
```
mzorkany, line 282: we can use the type `EntityResolutionServerStatus` instead of string

```suggestion
    this: Request<{status: EntityResolutionServerStatus}>,
```

src/api/entityResolution/index.ts
```diff
@@ -273,4 +274,25 @@ export class EntityResolutionAPI extends Request {
...
+    params: GetEntityResolutionServerStatusParams
+  ) {
+    return this.request({
+      errors: [
+        UNAUTHORIZED,
+        FORBIDDEN,
+        DATA_SOURCE_UNAVAILABLE,
+        NOT_FOUND,
+        ENTITY_RESOLUTION_EXPIRED_LICENSE
```
mzorkany, line 291: We also need to add the error `FEATURE_DISABLED`  in case the API is called and ER is not enabled
see:
https://github.com/Linkurious/linkurious-server/blob/faca79f88479f4763b51d36e7240ed06898cf068/server/routes/entityResolution.ts#L47-L48

aminerhanemi: I don't see FEATURE_DISABLED in any of the routes, I am not sure if we should add it only here

src/api/entityResolution/types.ts
aminerhanemi: In case we want to have more status, having an enum will make it easier (there is a low chance that it will happen, but just in case).

PR #841, opened 2025-06-26
LKE-13495: Disable log report download

gmasclet:
https://linkurious.atlassian.net/browse/LKE-13495


src/api/Config/types.ts
```diff
@@ -53,6 +53,7 @@ export interface Configuration {
   entityResolution?: EntityResolutionConfig;
   entityResolutionSourceKey?: string;
   cluster?: ClusterConfig;
+  troubleshooting: TroubleshootingConfig;
```
gmasclet, line 56: @EricMallard this property is actually only set for admin users, so it's optional. I'm going to push a fix

```suggestion
  troubleshooting?: TroubleshootingConfig;
```

PR #845, opened 2025-06-30
LKE-13320: Improve error handling

gmasclet:
https://linkurious.atlassian.net/browse/LKE-13320

### DIFF
--- a/src/http/response.ts
+++ b/src/http/response.ts
@@ -73,6 +73,7 @@ export enum LkErrorKey {
  * Basic error to be extended by custom errors
  */
 export interface LkError<K extends LkErrorKey = LkErrorKey> {
+  id?: string;
   key: K;
   message: string;
 }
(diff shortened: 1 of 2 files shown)


src/http/response.ts
```diff
@@ -73,6 +73,7 @@ export enum LkErrorKey {
  * Basic error to be extended by custom errors
  */
 export interface LkError<K extends LkErrorKey = LkErrorKey> {
+  id?: string;
```
gmasclet, line 76: The `LkError` instances returned by the server will always have an `id`. However, there is at least one case where the REST client itself throw a `LkError` instance:

https://github.com/Linkurious/linkurious-rest-client/blob/4536fb094d5985ffde7ed41437d8a3af66c54c20/src/http/request.ts#L180-L188

That's why the `id` field is optional. Alternative ideas are welcome:
* Changing the REST client itself so that it no longer throw this `LkError` is dodgy, it could be breaking (for plugins developers typically).
* Adding an ID to the `LkError` thrown by the REST client is confusing, as the ID won't be in the server logs (it's not a server error).

PR #920, opened 2026-01-19
LKE-14461: Enable ESLint cache and decouple coverage

gmasclet:
https://linkurious.atlassian.net/browse/LKE-14461


enys (approved):
I would have separated both PR's, as both have different concerns, and will not help measuring the gains. Approving nontheless, considering the short durations on rest-client.
Does this mean you are dropping coverage everywhere ?

PR #931, opened 2026-02-05
LKE-14494: Fix the ImportTemplate type

gmasclet:
Partially revert https://github.com/Linkurious/linkurious-rest-client/pull/930

### DIFF
--- a/src/api/import/types.ts
+++ b/src/api/import/types.ts
@@ -80,7 +80,8 @@ export interface GetImportTemplatesParams extends IDataSourceParams {
   entityType?: EntityType;
 }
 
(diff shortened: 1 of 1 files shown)


src/api/import/types.ts
```diff
@@ -80,7 +80,8 @@ export interface GetImportTemplatesParams extends IDataSourceParams {
   entityType?: EntityType;
 }
 
-export interface ImportTemplate extends UpdateImportTemplateParams {
```
gmasclet, line 83: Extending `UpdateImportTemplateParams` is wrong as it implies that most fields are optional. Hence the revert.

WilliamGorge (approved):

PR #939, opened 2026-03-25
LKE-14701: Add the import graph API

gmasclet:
## API Draft for https://linkurious.atlassian.net/browse/LKE-14701

* Add the option to populate a visualization sandbox by import ID. Used to add imported data to a new visualization.

* Add a new endpoint to get the imported graph for a given import ID. Used to add imported data to an existing visualization.

* Add `name` and `description` properties to the imports.

## Comments

* I assume that the frontend can use `GET /api/:sourceKey/visualizations/tree` to fetch visualizations for the dropdown list of the existing visualizations.

* We implement no limit / pagination on the new "get imported graph" endpoint or to populate a visualization sandbox by import ID. 
  * That might be an issue for large imports, albeit the actual limits aren't clear and may vary. Product is inform and OK with that. 
  * We'll display a warning banner in the UI: `Adding more than X entities at once is overwhelming. The process might take a few minutes.`. With X being 500 (constant - not configurable).

* In the backend, we'll need to add two new GraphDAO methods to get nodes/edges by property.
  * We implement these methods only on Neo4j and Memgraph.
  * We introduce a graph feature flag that completely hides the CSV import feature for other graph vendors.

* Legacy imports don't have a name. For these, the name will be an empty string.

* In the backend, the support for alternative IDs should be transparent.

### DIFF
--- a/src/api/import/index.ts
+++ b/src/api/import/index.ts
@@ -6,12 +6,14 @@
  */
 import {Request} from '../../http/request';
 import {LkErrorKey} from '../../http/response';
+import {LkSubGraph} from '../graphItemTypes';
 
 import {
   CreateImportParams,
   CreateImportTemplateParams,
   DeleteImportParams,
   DeleteImportTemplateParams,
+  getImportGraphParams,
   GetImportsParams,
   GetImportTemplatesParams,
   Import,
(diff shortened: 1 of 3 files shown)


WilliamGorge:
@gmasclet about the limit I fear 10K nodes/edges might be overwhelming for the frontend; I think we usually limit queries to 500, using the `rawQueryLimit` config, right?

src/api/import/index.ts
```diff
@@ -108,4 +110,16 @@ export class ImportAPI extends Request {
       params: params
     });
   }
+
+  /**
+   * Get the nodes and/or edges imported by a given import.
```
davidrapin, line 115: will the edges be returned with their extremities? if so, worth explaining that in the comment

davidrapin (approved):

gmasclet:
> @gmasclet about the limit I fear 10K nodes/edges might be overwhelming for the frontend; I think we usually limit queries to 500, using the `rawQueryLimit` config, right?

That's right. `searchAddAllThreshold` is defaults to 500 and the `expandThreshold` is actually 50.

Indeed, maybe 500 is more conservative. Or 1000. We'll have to benchmark that.

src/api/import/index.ts
gmasclet: for sure: the goal is to populate a viz, so we need both nodes and edges is any case. This is just a draft to illustrate the specs, so I won't polish the jsdoc yet, but it will be worth adding some details in the final version.

gmasclet:
Superseded by https://github.com/Linkurious/linkurious-rest-client/pull/940 and https://github.com/Linkurious/linkurious-rest-client/pull/941

PR #944, opened 2026-04-03
LKE-14101 replace superagent with native fetch

davidrapin:
bonus: bundle as a correct ESM module

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "@types/express": "4.17.15",
     "@types/mocha": "10.0.6",
     "@types/node": "24.12.0",
-    "@types/superagent": "4.1.20",
     "@typescript-eslint/eslint-plugin": "7.15.0",
     "@typescript-eslint/parser": "7.15.0",
     "esbuild": "0.28.0",
@@ -50,12 +49,12 @@
     "eslint-plugin-es5": "1.5.0",
     "eslint-plugin-import": "2.29.0",
     "eslint-plugin-prettier": "5.0.1",
+    "fetch-cookie": "3.2.0",
     "mocha": "10.2.0",
     "mocha-junit-reporter": "2.2.1",
     "mocha-multi-reporters": "1.5.1",
     "nyc": "15.1.0",
     "prettier": "3.1.1",
-    "superagent": "8.1.2",
     "typescript": "5.5.4"
   },
   "publishConfig": {
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -17,13 +17,7 @@ import {
   LkErrorKeyToInterface,
   Response
 } from './response';
-import {
-  FetchConfig,
-  ModuleProps,
-  RawFetchConfig,
-  SendBeaconConfig,
-  SuperAgentResponse
-} from './types';
+import {FetchConfig, ModuleProps, RawFetchConfig, SendBeaconConfig} from './types';
 
 export abstract class Request<S = undefined> {
   constructor(public readonly props: ModuleProps) {}
@@ -154,11 +148,11 @@ export abstract class Request<S = undefined> {
       requiredConfig = Request.renderURL(rawFetchConfig, this.props);
     } catch (error) {
       if (this.isDataSourceUnavailableError(error)) {
-        // 1.a) Return this when currentSource is not connected without performing an HTTP request
+        // Return this when currentSource is not connected without performing an HTTP request
         this.props.dispatchError(error.key, error);
         return new Response({body: error}) as ErrorResponses<EK>;
       } else {
-        // 1.b) Throw an exception when path params are missing
+        // Throw an exception when path params are missing
         throw error;
       }
     }
@@ -167,38 +161,33 @@ export abstract class Request<S = undefined> {
     const fetchConfig = Request.splitParams(requiredConfig, this.props);
 
     // 3) Make the HTTP request
-    let response: SuperAgentResponse;
+    let response: Response<unknown>;
     try {
-      response = await this.props.agent[
-        fetchConfig.method.toLowerCase() as 'get' | 'delete' | 'post' | 'put' | 'patch'
-      ](fetchConfig.url)
-        .ok((res) => res.status < 500)
-        .withCredentials()
-        .send(fetchConfig.body)
-        .query(fetchConfig.query);
+      response = await this.doRequest(fetchConfig);
     } catch (ex) {
-      // 4.a) Return error when there is no connection
-      if (!this.hasResponse(ex)) {
-        const error: ConnectionRefusedError = {
-          key: LkErrorKey.CONNECTION_REFUSED,
-          message: 'offline',
-          fetchConfig: fetchConfig
-        };
-        this.props.dispatchError(error.key, error);
-        return new Response({body: error});
-      }
+      const error: ConnectionRefusedError = {
+        key: LkErrorKey.CONNECTION_REFUSED,
+        message: 'offline',
+        fetchConfig: fetchConfig
+      };
+      this.props.dispatchError(error.key, error);
+      return new Response({body: error});
+    }
 
-      // 4.b) Throw error if status code is 5xx
-      throw new InternalServerError(ex.response);
+    // Throw error if status code is 5xx
+    if (response.status >= 500) {
+      if (!response.body) {
+        response.body = {
+          key: LkErrorKey.BUG,
+          message: 'Missing response body'
+        };
+      }
+      throw new InternalServerError(response);
     }
 
     // From here we only deal with responses with status code lower than 500
     if (this.isLkError(response.body)) {
-      const errorResponse = new Response({
-        status: response.status,
-        header: response.header as unknown as GenericObject | undefined,
-        body: response.body as LkErrorKeyToInterface[LkErrorKey]
-      }) as ErrorResponses<EK>;
+      const errorResponse = response as ErrorResponses<EK>;
 
       if (includes(requiredConfig.errors, response.body.key)) {
         // Dispatch server error if expected
@@ -225,11 +214,11 @@ export abstract class Request<S = undefined> {
     return new Response({
       status: response.status,
       header: response.header as unknown as GenericObject | undefined,
-      body: response.body as unknown as S
+      body: response.body as S
     });
   }
 
-  private isLkError(body: unknown): body is LkError {
+  public isLkError(body: unknown): body is LkError {
     return (
       body !== null &&
       typeof body === 'object' &&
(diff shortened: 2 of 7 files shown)


src/http/request.ts
```diff
@@ -247,7 +236,58 @@ export abstract class Request<S = undefined> {
     );
   }
 
-  private hasResponse<B>(error: unknown): error is {response: Response<B>} {
-    return (error as {response: Response<B>}).response !== undefined;
+  private async doRequest<T>(fetchConfig: FetchConfig): Promise<Response<T>> {
```
davidrapin, line 239: 99% of the change lives here

src/http/request.ts
```diff
@@ -167,38 +161,33 @@ export abstract class Request<S = undefined> {
...
-        .query(fetchConfig.query);
+      response = await this.doRequest(fetchConfig);
     } catch (ex) {
-      // 4.a) Return error when there is no connection
-      if (!this.hasResponse(ex)) {
-        const error: ConnectionRefusedError = {
-          key: LkErrorKey.CONNECTION_REFUSED,
-          message: 'offline',
-          fetchConfig: fetchConfig
+      const error: ConnectionRefusedError = {
```
davidrapin, line 168: 1% of the change lives here

src/http/request.ts
```diff
@@ -167,38 +161,33 @@ export abstract class Request<S = undefined> {
...
+        key: LkErrorKey.CONNECTION_REFUSED,
+        message: 'offline',
+        fetchConfig: fetchConfig
+      };
+      this.props.dispatchError(error.key, error);
+      return new Response({body: error});
```
davidrapin, line 174: emulating the previous behavior: "connection refused" is returned, not thrown

src/http/request.ts
```diff
@@ -167,38 +161,33 @@ export abstract class Request<S = undefined> {
...
+          key: LkErrorKey.BUG,
+          message: 'Missing response body'
         };
-        this.props.dispatchError(error.key, error);
-        return new Response({body: error});
       }
-
-      // 4.b) Throw error if status code is 5xx
-      throw new InternalServerError(ex.response);
+      throw new InternalServerError(response);
```
davidrapin, line 185: reproducing the current behavior: 500 errors are thrown

src/http/request.ts
```diff
@@ -247,7 +236,58 @@ export abstract class Request<S = undefined> {
...
+        }
+        if (Array.isArray(value)) {
+          for (const v of value) {
+            urlWithQueryString.searchParams.append(key, String(v));
+          }
+        } else {
+          urlWithQueryString.searchParams.append(key, String(value));
+        }
+      }
+    }
```
davidrapin, line 249: formatting querystring params as a good client would (omit missing values, spread arrays)

src/http/request.ts
```diff
@@ -247,7 +236,58 @@ export abstract class Request<S = undefined> {
...
+    const fetchResponse = await this.props.fetchMethod(urlWithQueryString, {
+      // important: use uppercase methods, we had test failures when using lowercase "patch"
+      // see https://github.com/nodejs/undici/issues/1805#issuecomment-1344797706
+      method: fetchConfig.method.toUpperCase(),
+      headers: {
+        ...(fetchConfig.body ? {'Content-Type': 'application/json'} : {}),
+        ...this.props.customHeaders
+      },
+      credentials: 'include',
+      body: fetchConfig.body ? JSON.stringify(fetchConfig.body) : undefined
```
davidrapin, line 260: serialize the body to json ourselves

src/http/request.ts
```diff
@@ -247,7 +236,58 @@ export abstract class Request<S = undefined> {
...
+    // Normalize the body:
+    // - If the content type is JSON, parse it as JSON.
+    // - Otherwise, return the body as an ArrayBuffer (this is the case for binary responses such as Excel extracts).
+    const gotJson =
+      (fetchResponse.headers.get('content-type') ?? 'unknown')
+        .toLowerCase()
+        .indexOf('application/json') === 0;
+    const body = gotJson
+      ? ((await fetchResponse.json().catch(() => undefined)) as unknown)
+      : await fetchResponse.arrayBuffer();
```
davidrapin, line 278: if the response is not JSON, return an arrayBuffer, as superagent was doing (used for excel export)

src/index.ts
```diff
@@ -70,18 +70,20 @@ export class RestClient extends ErrorListener {
 
     this.clientState = {};
 
-    let agent = request.agent();
-    for (const [field, value] of options?.headers ?? []) {
-      agent = agent.set(field, value);
+    let fetchMethod = fetch;
+    if (!('window' in globalThis)) {
```
davidrapin: this is how we detect that we are in nodejs

package.json
```diff
@@ -22,41 +25,40 @@
...
     "mocha-multi-reporters": "1.5.1",
     "nyc": "15.1.0",
     "prettier": "3.1.1",
+    "ts-node": "10.9.2",
     "typescript": "5.5.4"
   },
-  "dependencies": {
-    "superagent": "8.1.2"
-  },
+  "dependencies": {},
```
davidrapin: no more deps, we generate a bundle with all deps included

tsconfig.json
```diff
@@ -1,20 +1,26 @@
 {
   "compilerOptions": {
     "declaration": true,
-    "module": "commonjs",
-    "moduleResolution": "node",
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
```
davidrapin: making an esm module

src/http/request.ts
```diff
@@ -247,7 +236,58 @@ export abstract class Request<S = undefined> {
...
+
+    const normalizedResponse = new Response<unknown>({
+      status: fetchResponse.status,
+      header: normalizedHeaders,
+      body: body
+    });
+
+    return normalizedResponse as Response<T>;
```
WilliamGorge:
```suggestion
    const normalizedResponse = new Response<T>({
      status: fetchResponse.status,
      header: normalizedHeaders,
      body: body as T
    });

    return normalizedResponse;
```

src/http/request.ts
```diff
@@ -247,7 +236,58 @@ export abstract class Request<S = undefined> {
     );
   }
 
-  private hasResponse<B>(error: unknown): error is {response: Response<B>} {
-    return (error as {response: Response<B>}).response !== undefined;
+  private async doRequest<T>(fetchConfig: FetchConfig): Promise<Response<T>> {
+    const urlWithQueryString = new URL(fetchConfig.url);
+    if (fetchConfig.query) {
```
WilliamGorge: typings says that `fetchConfig.query` is always defined, should we update them? or remove this if? or add an assertion?

src/http/request.ts
WilliamGorge: simplification proposal to avoid nested if/elses

```suggestion
      for (const [key, value] of Object.entries(fetchConfig.query)) {
        if (value === undefined || value === null) {
          continue;
        }
        const valueAsArray = Array.isArray(value) ? value : [value];
        for (const v of valueAsArray) {
          urlWithQueryString.searchParams.append(key, String(v));
        }
      }
    }
```

src/http/request.ts
```diff
@@ -167,38 +161,33 @@ export abstract class Request<S = undefined> {
     const fetchConfig = Request.splitParams(requiredConfig, this.props);
 
     // 3) Make the HTTP request
-    let response: SuperAgentResponse;
+    let response: Response<unknown>;
```
WilliamGorge, line 164: to bad we have `unknown` here

src/api/AccessRight/index.ts
```diff
@@ -4,12 +4,12 @@
  * - Created on 2019-09-27.
  */
 
-import {Request} from '../../http/request';
-import {LkErrorKey} from '../../http/response';
+import {Request} from '../../http/request.js';
```
WilliamGorge: why are all those import changing? is it the esm module module change?

I find this surprising

WilliamGorge (changes requested):
Thanks, i made a few comment on the code itself; now it is hard to guess if there will be any regressions, but if all tests pass (unit, e2e, auto, ...) let's assume we're good!

It isn't a breaking change right?

src/api/AccessRight/index.ts
davidrapin: yes I had to change this because we're moving to esm (it's either that or rename all files to .mts)

davidrapin:
@WilliamGorge post-review done. 

This should not be a breaking change, because the RestClient has not changed on the surface (contructor parameters are the same). The only "surface" change is that `restClient.*.props.agent` does not exist anymore. In practice, this is not documented and only some hacky plugins were using it in the past, and should not use it anymore (worth a double-check with Fanta).

All tests are passing in the frontend and backend PRs:
- https://github.com/Linkurious/linkurious-server/pull/6147
- https://github.com/Linkurious/linkurious-frontend/pull/6868

tsconfig.json
gmasclet: I would rather expect the combo esnext/bundler for a bundled lib. Why choosing nodeNext?

Unless I'm wrong, with `esnext/bundler`, we don't have to add extensions to all the imports in the sources.

```suggestion
    "module": "esnext",
    "moduleResolution": "bundler",
```

tsconfig.json
```diff
@@ -1,20 +1,26 @@
...
     "outDir": "dist",
     "rootDir": ".",
     "inlineSourceMap": true,
     "noImplicitReturns": true,
     "incremental": true,
-    "lib": ["es5", "dom"],
-    "target": "es5",
+    "lib": ["es2023", "dom"],
+    "target": "es2020",
     "noUnusedLocals": true,
```
gmasclet, line 13: Worth discussing why we choose this target. The code is ultimately bundled, so at first glance, I would expect `esnext`. Maybe also `esnext` lib, for the same reason.

tsconfig.json
```diff
@@ -1,20 +1,26 @@
...
     "strict": true,
     "noFallthroughCasesInSwitch": true,
     "allowUnreachableCode": false,
     "forceConsistentCasingInFileNames": true,
-    "skipLibCheck": true
-  }
+    "skipLibCheck": false,
```
gmasclet, line 18: We generally set `"skipLibCheck": true`. It's probably irrelevant for the REST client as there are no dependencies.

package.json
```diff
@@ -1,12 +1,15 @@
...
-    "dist"
+    "dist/index.mjs",
+    "dist/index.mjs.map",
+    "dist/**/*.d.ts"
   ],
-  "main": "./dist/index.js",
-  "module": "./dist/index.js",
+  "main": "./dist/index.mjs",
+  "type": "module",
+  "module": "./dist/index.mjs",
```
gmasclet: * `exports` is the modern alternative to `main`.
* `"type": "module"` is redundant if we explicitly bundle as a `.mjs` file.
* unsure if we need the `module` field.

package.json
```diff
@@ -22,41 +25,40 @@
   },
   "homepage": "https://github.com/Linkurious/linkurious-rest-client#readme",
   "scripts": {
-    "build": "tsc",
-    "test": "mocha dist/test",
-    "test:unit": "mocha --reporter mocha-multi-reporters --reporter-options configFile=.test-reporters.json dist/test",
-    "coverage": "nyc --reporter=text-summary --reporter=cobertura --report-dir=reports  mocha --reporter mocha-multi-reporters --reporter-options configFile=.test-reporters.json dist/test",
+    "build": "tsc && esbuild --bundle --sourcemap --target=es2020 --format=esm --outfile=dist/index.mjs index.ts",
+    "test": "mocha --require ts-node/register 'test/**/*.ts'",
+    "test:unit": "mocha --require ts-node/register --reporter mocha-multi-reporters --reporter-options configFile=.test-reporters.json 'test/**/*.ts'",
```
gmasclet: I strongly advise against using ts-node for tests. By experience:
* it does not scale as the test files are recompiled from scratch each time they are run (without leveraging ts incremental compilation).
* Code editor support is worse. In particular, typescript imports can no longer be automatically fixed.
* We no longer have linting. This is a issue in itself, lint and formatting issues are going to multiply over the long run (for an example, check the `linkurious-shared` which is affected by this issue).
* We're adding the `ts-node` dev dep.
* We're relying on the `ts-node` compilation to match the `tsc` one. Generally the case, but not guaranteed and can lead to very confusing issues when it's not the case.

A much better alternative is to compile the tests using `tsc`. I can help with that if needed.

package.json
```diff
@@ -22,41 +25,40 @@
...
+    "coverage": "nyc --reporter=text-summary --reporter=cobertura --report-dir=reports mocha --require ts-node/register --reporter mocha-multi-reporters --reporter-options configFile=.test-reporters.json 'test/**/*.ts'",
     "compile": "tsc -b",
-    "tsc": "tsc",
-    "lint": "eslint --fix --ext .ts .",
+    "lint": "tsc && eslint --fix --ext .ts .",
```
gmasclet: why `tsc` here?

gmasclet (changes requested):
I've focused on the esm/build changes. Looks like a great improvement, but worth discussing a lot of details.

WilliamGorge (approved):
LGTM on my review, i'll let you discuss Guillaume's review

gmasclet (approved):
I think we're now good to merge

PR #946, opened 2026-04-07
LKE-14101: Bundle as a ESM module

gmasclet:
Bundle as a ESM module.

Stems from https://github.com/Linkurious/linkurious-rest-client/pull/944. 

Compared to the original implementation, I've fixed a few issues. In particular:
* We no longer need no change all the imports in `src`.
* The tests do not rely on `ts-node`.


davidrapin (approved):
much cleaner than my approach, thank you!

PR #948, opened 2026-04-09
LKE-14101: Fix fetch binding in browser context

gmasclet:
Completes https://github.com/Linkurious/linkurious-rest-client/pull/944.

Ensure the fetch method is correctly bound in browser context.

### DIFF
--- a/src/restClient.ts
+++ b/src/restClient.ts
@@ -69,19 +69,15 @@ export class RestClient extends ErrorListener {
 
     this.clientState = {};
 
-    let fetchMethod = fetch;
-    if (!('window' in globalThis)) {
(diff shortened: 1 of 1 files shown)


src/restClient.ts
```diff
@@ -69,19 +69,15 @@ export class RestClient extends ErrorListener {
 
     this.clientState = {};
 
-    let fetchMethod = fetch;
```
gmasclet, line 72: if `fetch` is initialized as a method reference, it get bound to `this.props` at the call site. We get the following error:

```
'fetch' called on an object that does not implement interface Window.
```

davidrapin (approved):

PR #953, opened 2026-04-13
LKE-14802 Add data source vendor to DataSourceUserInfo

mzorkany:


davidrapin (approved):

gmasclet (changes requested):
A couple questions to discuss:

* Today, the frontend is already able to retrieve the graph vendor, via the config. Why do want to add this field also here?
**EDIT:** @davidrapin I see on Jira that you have proposed this because the config may sometimes be unavailable. In that case, we should change the frontend selectors that rely on the config as we likely have bugs today.
 
* Let's admit we really want to add this field here. Wouldn't we prefer to have `graphVendor` and `indexVendor`? `vendor` is confusing and will later be problematic if we need both vendors to be exposed.

mzorkany:
> A couple questions to discuss:
> 
> * Today, the frontend is already able to retrieve the graph vendor, via the config. Why do want to add this field also here?
>   **EDIT:** @davidrapin I see on Jira that you have proposed this because the config may sometimes be unavailable. In that case, we should change the frontend selectors that rely on the config as we likely have bugs today.
> * Let's admit we really want to add this field here. Wouldn't we prefer to have `graphVendor` and `indexVendor`? `vendor` is confusing and will later be problematic if we need both vendors to be exposed.

@gmasclet 
i am not sure i understood, regarding the vendor retrieval from the config, do you mean the config API ?

Also i dont understand what selectors do we need to change on the frontend as a result of adding the vendor property to the DataSourceUserInfo

For me its ok to change the property from `vendor` to `graphVendor` to avoid confusion

davidrapin:
@gmasclet 
- Today, the frontend can get the config of the current source *if you are an admin*. That's a big if. Non-admin users have currently no way to get the graph vendor of a data-source. Even when you are an admin, the sourceConfig.vendor field is sometimes a dynamic value, so cannot always be read via the config endpoint. 
- I agree that once we have shipped this, there will be an opportunity to improve code in the frontend based on this fueld, but that's out of scope for now
- I agree that "graphVendor" is a better name
 
What is your fear in terms of maintaining this new field?

gmasclet (approved):
@mzorkany @davidrapin Good for me as discussed together :+1:

PR #957, opened 2026-04-16
LKE-14824 fix relative urls

davidrapin:
@justinmarsan

### DIFF
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -245,8 +245,8 @@ export abstract class Request<S = undefined> {
     );
   }
 
-  private async doRequest<T>(fetchConfig: FetchConfig): Promise<Response<T>> {
-    const urlWithQueryString = new URL(fetchConfig.url);
+  private getUrlWithQueryString(fetchConfig: FetchConfig): string {
+    const urlWithQueryString = new URL(fetchConfig.url, globalThis.document?.baseURI);
     for (const [key, value] of Object.entries(fetchConfig.query)) {
       if (value === undefined || value === null) {
         continue;
@@ -256,6 +256,11 @@ export abstract class Request<S = undefined> {
         urlWithQueryString.searchParams.append(key, String(v));
       }
     }
+    return urlWithQueryString.toString();
+  }
+
+  private async doRequest<T>(fetchConfig: FetchConfig): Promise<Response<T>> {
+    const urlWithQueryString = this.getUrlWithQueryString(fetchConfig);
 
     const fetchResponse = await this.props.fetchMethod(urlWithQueryString, {
       // important: use uppercase methods, we had test failures when using lowercase "patch"
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -104,40 +104,74 @@ export abstract class Request<S = undefined> {
   }
 
   /**
-   * Sort `config.params` into `config.body` and `config.query`
-   * and set `guest` and `_` query params.
+   * - Split `config.params` into body and query-string parameters depending on the HTTP method.
+   * - Set `guest` and `_` query-string params.
+   * - Normalize query-string params to snake_case, as this is what the Linkurious server expects.
+   * - Construct the URL by merging `baseUrl`, `config.url` and the query-string parameters.
    */
-  public static splitParams(
+  public static getFetchConfig(
     config: Required<RawFetchConfig>,
-    moduleProps: ModuleProps
+    moduleProps: ModuleProps,
+    now: number = Date.now()
   ): FetchConfig {
-    // 1) Default values for `body` and `query`
+    // 1. Default values for `body` and `query`
     let body: GenericObject | undefined;
     let query: GenericObject = {
-      _: Date.now(),
+      _: now,
       guest: moduleProps.clientState.guestMode ? true : undefined
     };
 
-    // 2) Split params into `body` and `query` depending on the method
+    // 2. Split params into `body` and `query` depending on the method
     if (includes(['GET', 'DELETE'], config.method)) {
       query = {...query, ...config.params};
     } else {
       body = config.params;
     }
 
-    // 3) Return a valid fetch config
+    // 3. Normalize query-string params to snake_case, as this is what the Linkurious server expects
+    const normalizedQuery = Request.toSnakeCaseKeys(query);
+
+    // 4. Merge the query-string params into the URL
+    // this can throw is the constructed URL is invalid (e.g. if baseUrl is not a valid URL)
+    const url = this.getUrlWithQueryString(moduleProps.baseUrl + config.url, normalizedQuery);
+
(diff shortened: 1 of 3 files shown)


src/http/request.ts
```diff
@@ -245,8 +245,8 @@ export abstract class Request<S = undefined> {
     );
   }
 
-  private async doRequest<T>(fetchConfig: FetchConfig): Promise<Response<T>> {
-    const urlWithQueryString = new URL(fetchConfig.url);
+  private getUrlWithQueryString(fetchConfig: FetchConfig): string {
+    const urlWithQueryString = new URL(fetchConfig.url, globalThis.document?.baseURI);
```
davidrapin: this avoids this call from failing is fetchConfig.url is a relative URL

davidrapin (commented):
@gmasclet fyi

src/http/request.ts
```diff
@@ -104,49 +104,85 @@ export abstract class Request<S = undefined> {
...
   /**
-   * Sort `config.params` into `config.body` and `config.query`
-   * and set `guest` and `_` query params.
+   * - Split `config.params` into body and query-string parameters depending on the HTTP method.
+   * - Set `guest` and `_` query-string params.
+   * - Normalize query-string params to snake_case, as this is what the Linkurious server expects.
+   * - Construct the URL by merging `baseUrl`, `config.url` and the query-string parameters.
    */
-  public static splitParams(
+  public static getFetchConfig(
```
davidrapin, line 124: Renamed this method and moved all the logic for generating a correct FetchConfig here.
We no longer generate the final fetch url in a separate step and do it here, via the getUrlWithQueryString method.

src/http/request.ts
```diff
@@ -104,49 +104,85 @@ export abstract class Request<S = undefined> {
...
-      query: Request.toSnakeCaseKeys(query)
+      // important: make sure the methods is uppercase, we had errors when using lowercase "patch"
+      // see https://github.com/nodejs/undici/issues/1805#issuecomment-1344797706
+      method: config.method.toUpperCase() as FetchConfig['method'],
+      url: url,
+      body: body
     };
   }
 
+  private static getUrlWithQueryString(url: string, queryString: GenericObject): URL {
```
davidrapin, line 160: moved out of doRequest

src/http/request.ts
```diff
@@ -104,49 +104,85 @@ export abstract class Request<S = undefined> {
...
+    try {
+      const urlWithQueryString = new URL(url, globalThis.document?.baseURI);
```
davidrapin: main change to fix the bug: pass an optional second parameter to `new URL` which contains the `baseURI` of the current page.

- In node.js, `globalThis.document?.baseURI` is undefined.
- In a browser, this is the url of the current page, corrected by the content of the `<base>` tag (see https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURI)

test/restClient.spec.ts
```diff
@@ -10,74 +10,152 @@ import {describe, it} from 'mocha';
...
-      Request.renderURL({
-        url: '/visualizations/:visualizationId/share/:userId',
-        params: {
-          visualizationId: 1,
-          userId: 2
-        }
-      }).url,
-      '/visualizations/1/share/2'
-    );
+  describe('getFetchConfig', () => {
```
davidrapin, line 93: added tests

src/http/types.ts
```diff
@@ -32,10 +32,9 @@ export interface SendBeaconConfig<
 }
 
 export interface FetchConfig {
-  url: string;
+  url: URL;
   method: 'GET' | 'DELETE' | 'POST' | 'PUT' | 'PATCH';
   body?: GenericObject;
-  query: GenericObject;
```
davidrapin, line 38: removed, this is now directly part of the "url" field

src/http/request.ts
```diff
@@ -104,49 +104,85 @@ export abstract class Request<S = undefined> {
...
    */
-  public static splitParams(
+  public static getFetchConfig(
     config: Required<RawFetchConfig>,
-    moduleProps: ModuleProps
+    moduleProps: ModuleProps,
+    now: number = Date.now()
   ): FetchConfig {
-    // 1) Default values for `body` and `query`
+    // 1. Default values for `body` and `query`
```
gmasclet, line 129: Consider removing the numbering altogether:
* sooner or later, we rearrange the logic and the numbers becomes inconsistent.
* Maybe it's just me, but I don't really see the value of numbering comments.

```suggestion
    // Default values for `body` and `query`
```

src/http/request.ts
```diff
@@ -104,49 +104,85 @@ export abstract class Request<S = undefined> {
...
     } else {
       body = config.params;
     }
 
-    // 3) Return a valid fetch config
+    // 3. Normalize query-string params to snake_case, as this is what the Linkurious server expects
+    const normalizedQuery = Request.toSnakeCaseKeys(query);
+
+    // 4. Merge the query-string params into the URL
+    // this can throw is the constructed URL is invalid (e.g. if baseUrl is not a valid URL)
```
gmasclet:
```suggestion
    // this can throw if the constructed URL is invalid (e.g. if baseUrl is not a valid URL)
```

src/http/request.ts
```diff
@@ -104,49 +104,85 @@ export abstract class Request<S = undefined> {
...
+    const url = this.getUrlWithQueryString(moduleProps.baseUrl + config.url, normalizedQuery);
```
gmasclet, line 148: Is OK to concatenate using `moduleProps.baseUrl + config.url`?

I guess the baseUrl has never a trailing slash and the relative config url has always a leading slash. The question is, are we sure about that?

src/http/request.ts
```diff
@@ -104,49 +104,85 @@ export abstract class Request<S = undefined> {
...
+        const valueAsArray = Array.isArray(value) ? value : [value];
+        for (const v of valueAsArray) {
+          urlWithQueryString.searchParams.append(key, String(v));
+        }
+      }
+      return urlWithQueryString;
+    } catch (error) {
+      if (error instanceof TypeError) {
+        throw new Error(`Invalid URL: ${url}`);
+      }
```
gmasclet: Is this OK?

I would be more comfortable if we had the original `error` attached as the `cause` of the rethrown one.

gmasclet (approved):
Good for me, just a few minor comments :+1:

src/http/request.ts
davidrapin: this is was we always did: `url: moduleProps.baseUrl + config.url,`

src/http/request.ts
davidrapin: this is a better than what `new URL` throws

davidrapin (commented):
post review done

